PR #262, opened 2021-08-11
Fix/anon in concordance list

tcouch:


tcouch:
I've converted the concordance_list view to a class-based view called ConcordanceListView as suggested by @raquel-ucl

src/rard/research/views/concordance.py
```diff
@@ -1,41 +1,79 @@
...
+from django.shortcuts import get_object_or_404, render
 from django.utils.decorators import method_decorator
 from django.views.decorators.http import require_POST
-from django.views.generic import ListView
+from django.views.generic import View
 from django.views.generic.edit import CreateView, DeleteView, UpdateView
 
 from rard.research.models import (AnonymousFragment, Concordance, Fragment,
-                                  OriginalText)
+                                  OriginalText, original_text)
```
raquelalegre: That's the other one I had spotted that I couldn't find yesterday. What's that `original_text` and how is it different from `OriginalText`?

tcouch: I have no memory of adding that in, but I presume I must have. Anyway, I've removed it and all tests still pass.

raquel-ucl (approved):
Looks good!

PR #263, opened 2021-08-12
Added some notes on the way we work.

raquel-ucl:
And a link to Zenhub before I can't find it anymore :)


raquelalegre:
@tcouch can you add here some notes on using pre-commit?

tcouch:
@raquel-ucl I've updated the zenhub link in your section and added a section on pre-commit. That also led me to add something about installing local requirements.txt so developers get all the linters and pre-commit.

PR #271, opened 2021-09-03
Feature/complex folds

### LINKED ISSUE #20: Search substitutions
Specifically:
  * vul <-> vol
  * v <-> u
 
Antonino will provide a full list

tim-band:
It's a bit light on tests, and it's hard to see if it will actually make anybody happy.
One thing that seems clear to me is that we really need to search for terms separately (and quotes for phrases) before this is really going to be nice. Currently with this fix it can get really woolly.
Anyway, it works and isn't too slow (at least for any queries I tried against the May database)

### DIFF
--- a/src/rard/research/tests/views/test_search.py
+++ b/src/rard/research/tests/views/test_search.py
@@ -153,29 +153,29 @@ class TestSearchView(TestCase):
         # basic tests for antiquarian
         a1 = Antiquarian.objects.create(name='findme', re_code='1')
         a2 = Antiquarian.objects.create(name='foo', re_code='2')
-        self.assertEqual(list(view.antiquarian_search('findme')), [a1])
-        self.assertEqual(list(view.antiquarian_search('FinDMe')), [a1])
-        self.assertEqual(list(view.antiquarian_search('foo')), [a2])
-        self.assertEqual(list(view.antiquarian_search('fOo')), [a2])
-        self.assertEqual(list(view.antiquarian_search('F')), [a1, a2])
+        self.assertEqual(list(view.antiquarian_search(SearchView.Term('findme'))), [a1])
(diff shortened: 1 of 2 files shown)


raquelalegre:
Thanks for this @tim-band! It looks good and performance seems OK locally for me too. One thing that puzzles me is that if I search for "uarro" and "varro" I get a different number of results. Same for "Lucius" and "Lvcius". Am I missing something?

The other thing is the list of substitutions also includes some of these occur at the beginning or end of a word. I guess that adds another level of complexity we can deal with separately.

tim-band:
> Thanks for this @tim-band! It looks good and performance seems OK locally for me too. One thing that puzzles me is that if I search for "uarro" and "varro" I get a different number of results. Same for "Lucius" and "Lvcius". Am I missing something?

Some fields don't have folding applied, so that would be why. I only put it on fields that were likely to have a lot of Latin in them, so uarro won't find all of Varro's actual stuff, whereas varro will.

> The other thing is the list of substitutions also includes some of these occur at the beginning or end of a word. I guess that adds another level of complexity we can deal with separately.

Yeah, this is a bit of a pain. I'm not sure how the same mechanism can even work, as LIKE doesn't have features like "end of word", so we'd have to search for "um ", "um,", "um.", "um?", "um!" and so on... I think we'll have to live without it.

tim-band:
> Yeah, this is a bit of a pain. I'm not sure how the same mechanism can even work, as LIKE doesn't have features like "end of word", so we'd have to search for "um ", "um,", "um.", "um?", "um!" and so on... I think we'll have to live without it.

Actually, it looks like we have REGEXP_REPLACE that allows \m and \M for beginning and end of words.

tim-band:
PR#279 subsumes this

tim-band:
This is superseded by other PRs

PR #275, opened 2021-09-22
Feature/convert anonymous unlinked

### LINKED ISSUE #231: Convert an ‘Unlinked’ fragment to ‘Anonymous’ and vice versa
Strange as it sounds, we occasionally change our minds about whether a fragment is ‘anonymous’ or not (most commonly when we reconsider whether to label something as a proper fragment of an antiquarian, or as an anonymous ‘appositum’ to that antiquarian’s other material). However, anonymous fragments in the database are in their own special category: once a fragment has been created as ‘anonymous’ it can never be attributed to an antiquarian, as ‘unlinked’ fragments can; likewise, a fragment once attributed to somebody can never be made anonymous (if it has no attribution, it becomes ‘unlinked’).

Could anonymous fragments have the option to be made unlinked, and vice versa? Ideally, the edit page on an unlinked fragment could have a button ‘Make anonymous’, and the edit page on an anonymous fragment with no appositum links could have a button ‘Make unlinked’.

tcouch:
Closes #231


tim-band:
This looks great, but it needs test code.

src/rard/static/js/project.js
```diff
@@ -94,13 +94,17 @@ $('form').click(function(event) {
...
+// confirm delete/convert of objects when forms submitted
 $('body').on("submit", "form", function (e) {
     var $clicked= $(this).data('clicked');
     if ($clicked.hasClass('confirm-delete')) {
         let what = $clicked.data('what') || 'object';
         return confirm("Are you sure you want to delete this " + what + '? This cannot be undone.');
     }
+    if ($clicked.hasClass('confirm-convert')) {
+        let what = $clicked.data('what') || 'object';
+        return confirm("Are you sure you want to convert this " + what + '? This cannot be undone.');
```
raquelalegre: Is there a way to use `hasClass` to check if a fragment has apposita? That way the warning message can also include a note on what'll happen to the apposita after the conversion.

raquelalegre (approved):
Approved pending change in warning message.

src/rard/static/js/project.js
tcouch: image
If an anonymous fragment has appositum links, the warning now includes a note explaining these will also be converted.

Issue #36, opened 2020-12-01
Set up pre-prod VM


raquel-ucl:
This has now been requested by Wellesley.

Once it is ready, we'll have to:
* [x] Request SSH access for RITS team to HIS, including sudo permissions so we can run docker
* [x] Request firewall ports opened for external access to ns-core
* [ ] Request to create service id to Webops       
* [x] Email eligibility to request approval for url (www.ucl.ac.uk/history/frrant-dev?)
* [ ] Map url to pre-prod (request ns-core or platform integrations?)
* [ ] Request SSL certificate to Webops      
* [ ] Apply SSL cerificate  
* [ ] Request access for monitoring software and email alerts to HIS
* [ ] Set up UptimeRobot

tcouch:
Now accessible publicly at https://frra-pp-10-00_524.gtm.ucl.ac.uk/frrant/ 
To do:
- [x] Add a big yellow banner saying development or pre-production to the top of every page
- [ ] Sort out the certificate
- [x] Sort out the URL mapping to ucl.ac.uk/history/frrant-preprod

raquelalegre:
Now also accessible at https://www.ucl.ac.uk/history/frrant-preprod/, however all links take to the https://www.ucl.ac.uk/history/frrant version, which is a bit dangerous. Configuration needs changing as soon as possible.

raquelalegre:
Configuration has been changed and both prod and pre-prod versions have been live for 2 weeks.

PR #282, opened 2021-10-14
Fix/development server config

tcouch:
We have a new hostname for the pre-prod server (frra-pp-10-00_524.gtm.ucl.ac.uk) which may not be permanent, but if we want to access the pre-prod server via the web, we need to fix some issues:
* Add hostname to nginx config server_name
* Add hostname to the self-signed cert
* Monkey patch Django so it'll allow hostnames containing underscores (otherwise you'll get Bad Request 400)

Related to #36

### DIFF
--- a/src/compose/development/nginx/conf.d/rard.conf
+++ b/src/compose/development/nginx/conf.d/rard.conf
@@ -16,7 +16,7 @@ server {
         root /;
     }
 
-    location /frrant {
+    location /history/frrant {
         resolver 127.0.0.11;
         proxy_pass http://django;
         proxy_set_header Host $http_host;
@@ -49,7 +49,7 @@ server {
     }
 
     # Django for everything else
-    location /frrant/ {
+    location /history/frrant/ {
         resolver 127.0.0.11;
         proxy_pass http://django;
         proxy_set_header Host $http_host;
--- a/src/compose/development/nginx/conf.d/rard.conf
+++ b/src/compose/development/nginx/conf.d/rard.conf
@@ -6,7 +6,7 @@ upstream django {
 server {
     listen 80;
 
-    server_name frra-app01p.ad.ucl.ac.uk;
+    server_name frra-app01p.ad.ucl.ac.uk frra-pp-10-00_524.gtm.ucl.ac.uk;
 
     # Permanent HTTPS redirect
     #return 302 https://$server_name$request_uri;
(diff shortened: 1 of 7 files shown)


raquelalegre (commented):
That was quick! No need to change the URL_PREFIX const in ./.envs/.development/.django?

src/compose/development/nginx/conf.d/rard.conf
```diff
@@ -41,21 +40,27 @@ server {
...
-    location /frrant/ {
+    location / {
+        # Use docker DNS to resolve //django correctly
         resolver 127.0.0.11;
+        # Proxy /history/frrant-preprod/PATH to http://django/PATH
         proxy_pass http://django;
+        # Tell Django what the original request was for
         proxy_set_header Host $http_host;
-        proxy_set_header SCRIPT_NAME /frrant;
+        # Tell Django which bit of the path is the prefix
```
tim-band: This comment needs deleting

src/config/settings/development.py
```diff
@@ -38,6 +52,9 @@
 SESSION_COOKIE_SECURE = True
 # https://docs.djangoproject.com/en/dev/ref/settings/#csrf-cookie-secure
 CSRF_COOKIE_SECURE = True
+
+CSRF_TRUSTED_ORIGINS = ["www.ucl.ac.uk"]
```
tim-band, line 56: weird that frra-app01p.ad.ucl.ac.uk seems to be trusted too?

src/requirements/base.txt
```diff
@@ -8,7 +8,7 @@ hiredis==1.1.0  # https://github.com/redis/hiredis-py
 
 # Django
 # ------------------------------------------------------------------------------
-django==3.1  # https://www.djangoproject.com/
+django==3.2  # https://www.djangoproject.com/
```
tim-band, line 11: Didn't need this in the end (it was to make SCRIPT_NAME work, which still doesn't (at least in production), and actually it already worked in 3.1.6... Still, no harm in bumping it up?

tim-band (approved):

src/requirements/base.txt
tcouch: Don't see any reason why not

src/config/settings/development.py
tcouch: Yeah, I hadn't tested that one.

PR #283, opened 2021-10-26
Yellow nav bar and title label for local/dev

tcouch:
image
Instance var added to local.py, development.py, production.py settings files, accessible in templates as INSTANCE by expanding on pre-existing context processor.


tcouch:
As requested in #36

raquelalegre (changes requested):
Thanks for this! Looks good but boolean variable names shouldn't contain a negative form. So instead of a `non_prod` variable being true, I'd have a "prod" variable that is negative, or a "pre-prod" variable that is true. That'll make it easier to read.

PR #285, opened 2021-10-28
Confirm no data required for apparatus criticus

### LINKED ISSUE #179: Tick box to confirm no data needs to be entered in Apparatus Criticus
**Is your feature request related to a problem? Please describe.**
Difficult to tell whether there is no information to put in the Apparatus Criticus box (i.e. if box has been intentionally left blank), or if fragment has been created andd Apparatus still needs to be entered.

**Describe the solution you'd like**
A tick box to confirm whether the Apparatus Criticus has intentionally been left blank.

**Describe alternatives you've considered**

**Additional context**

tcouch:
closes #179


tcouch:
Add a warning instead of "-" where there are no app crits and checkbox not ticked.

tcouch:
Can we have a pop-up warning if people are leaving app crit blank but haven't ticked the box.

tcouch:
@raquel-ucl I'm happy with where this is up to now. Do you want to have a look or should I merge this now?

raquelalegre:
Having a look now.

raquelalegre:
I think there's something about the navigation when creating a fragment that is making this a little unintuitive.

So at the moment this is the workflow I follow:
1. From home, click on "Create fragment".
2. Fill in information without ticking app crit box and without adding any app crit.
3. Click on "Create".

I am redirected to a non editable view of the fragment that has a message for the apparatus criticus that reads "No apparatus criticus; confirmation needed.". That's fine to deploy to pre-prod and I'd rather wait for the team's feedback than follow mine, but I think it would be better if the fragment wasn't created until the user has declared if there's going to be an app crit or not. For example, instead of the checkbox, we could give the user two options: "Does this fragment have an app crit?" with choices "No" and "Yes, and I'll add it later". In the validation, one of these has to be chosen, otherwise the fragment is not created.

This problem is rooted in the slightly broken way we have to create fragments where you can't at the same time add links and create an app crit. If we ever have the time we should review and improve this workflow.

raquelalegre (approved):
Approving to deploy to pre-prod and see team's feedback.

tcouch:
@raquel-ucl my thinking was to give users the option of creating a fragment with the intention of adding the apparatus criticus later, and the warning message would make it clear to anyone viewing that fragment that it still needs work. This should  distinguish it from fragments which are complete without apparatus critici. I'll make that clear in my test instructions when I ask for feedback.

PR #292, opened 2021-11-10
Fix for the unlink button showing on anonymous fragment pages outside edit mode

### LINKED ISSUE #280: The "unlink" option in a fragment's apposita section should only display in edit mode.
At the moment it displays in view mode and that can be dangerous.

tcouch:
Closes #280 
Copying the way link items are displayed on Fragment detail pages, I've created a new partial template for anonymous fragment appositum link items. The anon frag detail page passes can_edit (does the user have permission) and has_object_lock to the partial, which only displays the unlink button if both are true.


raquelalegre (approved):
Looks good!

Issue #65, opened 2020-12-01
Search with *


raquelalegre:
Users should find "Varro" search results if they type:
* *rro
* Va*
* va*o

tcouch:
In addition from #297: "It would be extremely useful to have the ability to construct searches like pro*ur (to find proferretur), and also, if possible, something like profe??etur (where ? would stand for a single unknown character, thus only returning words of the required 'shape')."

tcouch:
Update from research meeting 04/04/2022: By default a keyword matches any part of a string. Can we add a tick box to make this an exact whole word match; e.g. "sed" matches "Amicus Plato, sed magis amica veritas" but not "possedere".

tcouch:
Above request reiterated in today's meeting with "ara" being another example; e.g. matching "ara maxima" and not apparatus

PR #297, opened 2021-11-18
Folded search and app crit confirmation

tcouch:
The two major features for review here are the new search functionality (automatic substitutions and phrase search) and adding checkbox confirmation for original texts which are not intended to have an apparatus criticus.


tcouch:
@rmamarshall I've added a fix for issue #179 adding a tick box to confirm no apparatus criticus need to be added. Some suggestions for things to try out on the pre production server follow:

# Create a Fragment
* From [home page](https://www.ucl.ac.uk/history/frrant-preprod/) click "create fragment"
* On Create Fragment page choose a Citing author and Citing work
* Fill in all fields relating to the original text. If at this stage you are sure the fragment you are creating will not include an apparatus criticus, you can tick the checkbox "no apparatus criticus exists".
* Either:
	* Click "Create" to create the new fragment and show its details page. You will need to unlock the fragment if you would like to make any further changes.
	* Click "Create and Add Links" to create the new fragment and show the form for adding links to an antiquarian, work or book.
	* Click "Create and Add Apparatus Criticus" to create the new fragment and show the original text update form which includes the apparatus criticus builder. You can also check/uncheck the box indicating there is no apparatus criticus on this page.
	
# View apparatus criticus
* On a fragment's details page (e.g. https://www.ucl.ac.uk/history/frrant-preprod/fragment/220) you will be able to see the apparatus criticus for all original texts associated with that fragment. This section will display either:
	* A list of all the apparatus criticus lines for the original text separated by a "|" character
	* The statement "No apparatus criticus exists for this original text" if no apparatus criticus lines exist and the checkbox confirming this has been checked.
	* A warning with the message "No apparatus criticus; confirmation needed" if no apparatus criticus lines have been added, and the checkbox confirming there is no apparatus criticus has not been checked.
	
# Update apparatus criticus
* Put the fragment in edit mode by clicking the Edit button on a fragment's details page
* At the top of the original text section, click Edit Details to show the Update Original Text Details Form
* Use the Apparatus Criticus Builder at the bottom of the page to add/update/remove apparatus criticus lines or use the checkbox to confirm there is no apparatus criticus
* **Note** the checkbox will be hidden once an apparatus criticus line has been added. If previously checked, it will be unchecked before being hidden. Remove all apparatus criticus lines to show the checkbox again.

tcouch:
@rmamarshall: Regarding the new search functionality. Search substitutions are in place for fields we'd expect to contain latin text; e.g. a search for "Clodius" will return fragments containing "Claudius" if it appears in the original text content, apparatus criticus or commentary; it won't show up if it only appears in the translation though. See the original issue #20 for a list of substitutions.

If a search term forms part of a word, it will still match; e.g. "Claud" will match Claudius, Claudii, Claudio, Claudium..., "ferret" will match proferret, adferretur etc. Regarding #65 (search with \*), is this sufficient, or do you envisage needing to use search terms like Poll*a and have it match Pollentia, Pollusca?

When searching for multiple terms separated by spaces, only items containing all those terms are returned (logical And rather than Or).

Tim's also added the ability to search for phrases by putting the term inside quotation marks; e.g. searching for "Arcadian sacrifice" only matches fragments containing the exact phrase and not fragments which happen to include both words non-consecutively.

rmamarshall:
@tcouch @raquel-ucl Thank you so much for this; will give it a spin and report ASAP.

rmamarshall:
@tcouch @raquel-ucl I have made a thorough trial of the new apparatus criticus feature (I've followed the above instructions, added text, deleted text, used the add original texts feature, tried different ways of creating a fragment etc.) and it seems to work brilliantly, thank you!

rmamarshall:
@tcouch @raquel-ucl 
I have a couple of queries regarding the search functionality:

Does punctuation influence search results? e.g. would a search for ferret return fer[ret ...]? It would be useful if [] and <> could be ignored (though I've no idea if this is even possible).

I presume Claud will also find clodius etc.?

Am I right in thinking that a search for ferret will act like *ferret* (returning e.g. proferretur), and that a search for "ferret" in inverted commas will only return ferret itself? To make searches of the sort ferret* rather than *ferret* (i.e. to ensure that only ferretur will be returned, not proferretur) or *ferret (for proferret, not proferretur), can a space be entered before / after the search term?

It would be extremely useful to have the ability to construct searches like pro*ur (to find proferretur), and also, if possible, something like profe??etur (where ? would stand for a single unknown character, thus only returning words of the required 'shape').

Another useful feature would be the ability to run proximity searches (i.e. find occurrences of two or more words within a certain number of words / characters of one another; I think PHI Latin do this using the ~ symbol, e.g. sed~et finds instances of these words within a set number of characters [and in that order]; I'm not sure if this works with * as well). It would also be great to have the ability to restrict searches to a particular antiquarian or citing author (e.g. search only Varro or Aulus Gellus).
 
Finally, it would be extremely useful to see snippets of text in the search results page (say three lines around the discovered term), rather than have to click through to each fragment. This feature in particular is one that will have a big effect on productivity / the usability of the database. Here is an example of what I have in mind:

image

I realize I'm asking a lot and have no idea how complex the above might be to institute; if some features are easier to implement that others, I'm sure we can live without them! 

The suggested searches using logical And and "exact phrase" will be very useful - thank you!!

tcouch:
@rmamarshall in response to your question about punctuation, I wasn't sure so I tried a little experiment. See https://www.ucl.ac.uk/history/frrant-preprod/fragment/59/ where I've added "find[me]" to the original text. A search for "findme" **does** match that fragment, and "find[me]" also returns the same fragment. Is that what you'd expect?

tcouch:
@rmamarshall Thanks for your feedback. I'm going to merge this pull request now as I don't think there was anything you wanted us to fix with these specific changes (do correct me if I'm wrong). I've created new issues for the additional search-related features you asked about and perhaps we can discuss priorities for these at the next group meeting.

rmamarshall:
@tcouch Re: searching for find[me] - yes, that's just what we need: thank you! 
A test with round brackets, "find(them)", also works.
However, it does not seem to work for find<you>. Can this be fixed, or is it something to do with tagging?

tcouch:
@rmamarshall this was due to the < and > character being converted to their character code equivalents \&lt; and \&gt; when stored in the database. I've created a fix for this, which I'll add once I've got it reviewed.

PR #302, opened 2021-11-22
add query func to strip <> character codes

tcouch:
In #297 @rmamarshall pointed out angle brackets are not being ignored in search in the same way that other brackets are. This was due to <> being converted to their html character codes, "&lt;" and "&gt;", which were not then picked up by the translate function which strips out other forms of punctuation during the search.

I've added an additional query function to the basic query which uses postgres's regexp_replace method to remove these character codes prior to the rest of the punctuation characters being removed.


tcouch:
@tim-band I thought you might be the best person to review this since you wrote the original code. Hope that's ok.

Issue #289, opened 2021-11-05
Unlinked fragments do not show apposita on details page


tcouch:
Unlinking the antiquarian from a fragment turns it into and unlinked fragment, but this loses it's apposita - can we investigate if this is permanent and fix?

tcouch:
The problem is a more general one than first stated. Fragment detail pages list apposita by iterating through the collection of appositumfragmentlinks, but these links are not created when apposita are added to an unlinked fragment. Changing the details template to show fragment.apposita if it's an unlinked fragment should fix this.

PR #303, opened 2021-11-24
Fix/display unlinked apposita

tcouch:
Making an anonymous fragment an appositum to an unlinked fragment does no create a fragment link object. The fragment and anonymous fragment detail pages weren't taking this into account as they were just looping through all fragment link objects to display apposita. This PR fixes both detail pages to include unlinked<->anonymous fragment apposita.
Related to #289


raquelalegre (approved):
Looks good!

Just tested locally by editing an unlinked fragment to add an anon fragment as appositum. This relationship now shows both in the unlinked and anon fragment detail views.

Issue #160, opened 2021-02-10
Fragments in the work section need to be ordered by book


raquel-ucl:
@apitta0188 reported:

> When fragments are displayed in the Work section, they are not ordered by books [i.e. Varro’s fragments from Book 2 of De uita appear before some fragments from Book 1 of the same work]

PR #304, opened 2021-11-24
Work fragments ordered by book

tcouch:
Fix for #160 
I've added a get_context_data function to the WorkDetailView which passes a nested dictionary containing the work's books, their fragments, testimonia, and apposita, and separate lists for definite and possible links.
The work detail template loops through these dictionaries to display associated material ordered by book, then by material type (fragment, testimonium, appositum), then by definite/possible.
I've also added anchor links to help navigate from book titles in the details section to the area of the page listing that book's contents.

### DIFF
--- a/src/rard/research/models/work.py
+++ b/src/rard/research/models/work.py
@@ -121,3 +121,6 @@ class Book(HistoryModelMixin, DatedModel, BaseModel):
     def get_absolute_url(self):
         # link to its owning work
         return reverse("work:detail", kwargs={"pk": self.work.pk})
+
+    def get_display_string(self):
+        return self.__str__()
--- a/src/rard/research/models/work.py
+++ b/src/rard/research/models/work.py
@@ -1,6 +1,7 @@
 from django.contrib.postgres.aggregates import StringAgg
 from django.db import models
 from django.urls import reverse
+from django.utils.text import slugify
 from simple_history.models import HistoricalRecords
 
 from rard.research.models.mixins import HistoryModelMixin
@@ -122,5 +123,8 @@ class Book(HistoryModelMixin, DatedModel, BaseModel):
         # link to its owning work
         return reverse("work:detail", kwargs={"pk": self.work.pk})
 
-    def get_display_string(self):
+    def get_display_name(self):
         return self.__str__()
+
+    def get_anchor_id(self):
+        return slugify(self.__str__())
--- a/src/rard/research/views/work.py
+++ b/src/rard/research/views/work.py
@@ -42,11 +42,11 @@ class WorkDetailView(
         books = work.book_set.all()
         # Empty structure with space for materials with unknown book
         ordered_materials = {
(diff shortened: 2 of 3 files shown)


tcouch:
Works <siteurl>/work/37/ and <siteurl>/work/27/ give a good representation of this.

raquelalegre (approved):
Looks good! I think it can be merged to development to be deployed and tested in pre-prod, but I have some questions, see below.

This is what I tested locally:
* Varro's De vita populi romani already had many fragments assigned to books, and now they show grouped correctly, and those unassigned are under the "Unknown Book" section.
* Each book section also shows definite fragments, possible apposita, etc. 
* I changed book attributions and fragments show under expected sections.
* I deleted a book and fragments assigned to that book moved to the unknown book section.

3 questions:
* What's the order fragments follow within each book section? My understanding was the ordering had to be by fragment ID, but I might be mistaken. This is what I see for my local DB with Varro's _Antiquitates rerum humanarum_:

    varro_order
    
    The order in the unknown book section seems fine for all works I've checked.
    
* Should book sections also appear in the antiquarian's detail view? Something that we need to ask the team during testing perhaps.
* I get this error (it might have been here for a long time, but I hadn't noticed as the website is deployed and seems to work!):
    ```
    Exception occurred:
    File "/usr/local/lib/python3.8/site-packages/docutils/writers/html5_polyglot/__init__.py", line 445, in section_title_tags
    if (ids and self.settings.section_self_link
    AttributeError: 'Values' object has no attribute 'section_self_link'
    ```

Note branch names should be preceded by categories like `feature/`, `fix/` or `docs/`.

raquel-ucl:
Fixes #160

tcouch:
@raquel-ucl 

> What's the order fragments follow within each book section?

I haven't explicitly done anything to control the order. I think it makes sense to order wrt the antiquarian so I'll try to add that before merging

> I get this error...

Where/when do you get that error?

raquelalegre:
Would ordering by antiquarian change anything? If we are ordering inside a book, that belong to a work, that belongs to an antiquarian... I guess all the fragments would be by the same one. No? 

I get that error when I deploy the website with docker. I think it might just be a documentation issue, and we are not using those docs for anything right now. I just don't know when this was introduced. Do you not get it?

raquelalegre:
I think order needs to be by fragment ID. At least that's how it used to show before introducing this change. Maybe better ask the team before doing anything.

tcouch:
Fragment links have an order and work_order properties; this is where the numbers comes from in the name; e.g. Marcus Terentius Varro: De sermone Latino ad Marcellum F3 [= Marcus Terentius Varro F112] is the 112th with respect to Marcus Varro and 3rd with respect to the book.

The latest commit uses the order property to order the materials (after ordering by book and by definite/possible).

raquelalegre:
Great! It now shows as:

varro_order

Thanks @tcouch !

PR #305, opened 2021-11-24
Feature/include unlinked in mentions

tcouch:
Addressing #295, unlinked fragments can now be added via @ mentions. This works in a similar fashion to anonymous fragments search in that a keyword consisting of the letter "u" will find all unlinked fragments, and "u\<int\>" will find all unlinked fragments whose id starts with that number; e.g. "u33" will find Unlinked 338 and Unlinked 339.


raquelalegre (approved):
Great! 

Tested locally:
* typing "@" I can see the list of possibilities includes unlinked fragments too, although it takes a while to load!
* typing "@u361" it filters down as I type every character to the one I want to link to.

Issue #300, opened 2021-11-18
Restrict search to specific antiquarian or author


tcouch:
From #297 "It would also be great to have the ability to restrict searches to a particular antiquarian or citing author (e.g. search only Varro or Aulus Gellus)."

Issue #298, opened 2021-11-18
Show text snippets in search results


tcouch:
Responding to #297 @rmamarshall says "it would be extremely useful to see snippets of text in the search results page (say three lines around the discovered term), rather than have to click through to each fragment. This feature in particular is one that will have a big effect on productivity / the usability of the database."

Example:
https://user-images.githubusercontent.com/75932472/142429526-18c667ab-e761-4c8e-97e6-f21f0ab0c25e.png

tcouch:
Possible with django postgres fulltext search. After adding django.contrib.postgres to INSTALLED_APPS, we can run the following:
```
from django.contrib.postgres.search import SearchQuery, SearchHeadline

from rard.research.models import OriginalText

query = SearchQuery('Homeri maxime')

matches = OriginalText.objects.annotate(
    headline=SearchHeadline(
        'content',
        query,
        start_sel="<span>",
        stop_sel="</span>",
        max_fragments=5,
        max_words=7,
        min_words=3),
).filter(content__search=query)

for ot in result: print("\n",ot.headline)
```
Which will give us the following output:
```

 patria quoque <span>Homeri</span> multo <span>maxime</span> dissensum ... imaginibus primo <span>Homeri</span> imagini epigramma hoc apposuit

 super aetate <span>Homeri</span> atque Hesiodi non consentitur ... itidem" inquit "uel <span>maxime</span> quod unoculus fuit

 super aetate <span>Homeri</span> atque Hesiodi non consentitur ... itidem  " inquit "  uel <span>maxime</span> quod unoculus fuit
```

tcouch:
I have a proof of concept working, but there's a lot of work to do to get this working with the other search features that have been developed:
<img width="581" alt="image" src="https://user-images.githubusercontent.com/5113832/144826879-01fbd8a0-c1c8-4dc9-9f61-403442f1f095.png">

PR #314, opened 2022-01-17
Faceted search with snippets

tcouch:
New features adding preview snippets to search results and search filtering by antiquarian and author.
Addresses #300 and #298


tcouch:
@raquel-ucl I think I've fixed this issue now. If you still have problems I'd suggest doing the following to reset everything:
```
# bring all the containers down and remove the volumes to delete the database data
docker-compose -f local.yml down -v
# Make sure you're on the development branch and rebuild the containers
docker-compose -f local.yml up -d --build
# Load a recent database dump
./loaddata.sh dump.json
# Switch to feature/faceted-search branch
# Take everything down and bring it back up again (this will rerun migrations)
docker-compose -f local.yml down
docker-compose -f local.yml up -d
```

raquelalegre:
@tcouch This looks **great** and it works perfectly ⭐

raquelalegre (approved):
We just had a call to clarify some of my questions, like the magic behind the nested Func calls in `search.py`, and considerations in the future in case we introduce more text items in the DB which will might to have a `make_plain_text` call. All seems good. Only worry is performance now that there's a bit more "meat" to the post-processing of search results, but we'll benchmark it and see what improvements we can make once we deploy to development.

PR #316, opened 2022-02-23
Search snippets and filters

### LINKED ISSUE #298: Show text snippets in search results

### LINKED ISSUE #300: Restrict search to specific antiquarian or author

tcouch:
Two key changes:

1. Search snippets
2. Filter by Antiquarian/Citing Author

closes #300 
closes #298

# Search snippets
Search results now include short snippets of text from the larger body of text where the target word was found. The snippet starts 5 words before the target word and ends 5 words after. Where the target word has been found multiple times in the text, snippets will be shown for each, separated by "...". Snippets are not provided where the target content type is a short piece of text such as an antiquarian's name, or the tile of a work.

# Filter by Antiquarian/Citing Author
It is now possible to filter search results by Antiquarian or Citing Author. You can select a combination of antiquarians or citing authors from the entire database before searching. Or, following an unrestricted keyword search, the dropdown lists will just contain those antiquarians and citing authors associated with the returned search results. This would allow you to further narrow down the results by re-submitting the search with a subset of those antiquarians/citing authors selected.

The content type drop-down now also allows multiple content types to be selected.


rmamarshall:
@tcouch This looks brilliant!
I just have a couple of queries:
1) When I search for 'sed' or '_sed' (i.e. space followed by sed), the results return words like 'possedere'. I thought adding a space was supposed to return only words beginning 'sed-'.
2) When I search for 'sed', I also get hits from translations as well as original text; is there any way to filter these out?

tcouch:
Hi @rmamarshall. I'm afraid, at the moment the search doesn't allow you to specify that you want the term to only match the start of the word, but perhaps we could introduce a syntax for this. I think this is a similar type of problem to the wildcard issue #65 we've discussed before, but with additional syntax to denote the start and end of words. Perhaps when an asterisk is used in a search term, the positions of the other characters are fixed; e.g.:

- "*sed**" would find words starting with *sed*
- "**sed*" would find words ending with *sed*
- "*\*sed\**" would find words with *sed* in the middle
- "*sed*" would work as it does now, matching all words containing *sed* irrespective of the position

For the second issue you mentioned, one approach I can think of would be to break the content types filter down into more specific options; e.g.:
* Fragments - original text
* Fragments - translations
* Fragments - commentary

It'd mean a bit more clicking if you wanted to search across say original texts for apposita, fragments and testimonia, but it's the most straightforward option I can think of from a development point of view.

I'll have a think about how much work this would require so we can discuss it at the next group meeting, but in the meantime are you happy for me to push the current changes into production and create new issues for these features?

rmamarshall:
Hi @tcouch, so sorry, notifications are still going to an email address that I don't check very regularly any more. Yes, happy for the updates to be implemented (if this has not already been done) and for new issues to be created.

raquelalegre:
@rmamarshall if you go to https://github.com/settings/emails you can update your email preferences and add your UCL one so you get GitHub notifications there instead.

rmamarshall:
@raquel-ucl @tcouch Email address updated and notifications now going to my university account. Apologies once again for the delay in replying to Tom's last message; looking forward to seeing you this afternoon!

PR #328, opened 2022-06-07
Add 03BD to symbols.json

### LINKED ISSUE #327: Add symbol 03BD to symbol picker
@rmamarshall asks "May we add a missing letter to the Greek section of the symbol picker? This appears on p. 5 of the [alphabetum character set]"

Character code is 03BD

tcouch:
closes #327

### DIFF
--- a/src/.pre-commit-config.yaml
+++ b/src/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: check-yaml
 
   - repo: https://github.com/psf/black
-    rev: 21.8b0
+    rev: 22.3.0
     hooks:
       - id: black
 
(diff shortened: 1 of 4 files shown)


src/.pre-commit-config.yaml
```diff
@@ -11,7 +11,7 @@ repos:
       - id: check-yaml
 
   - repo: https://github.com/psf/black
-    rev: 21.8b0
```
raquelalegre, line 14: Could this be to blame for the tests not passing anymore?

raquelalegre (approved):
The code looks OK and the new symbol appears in the Greek section of the symbol picker between 03BC and 03BE.

src/.pre-commit-config.yaml
raquelalegre: Ah! I just notice my comment is outdated and that the tests are passing now. Plus this commit was actually what fixed precommit! 0:)

PR #329, opened 2022-06-21
Fix/make anonymous order

### LINKED ISSUE #309: ‘Make anonymous’-feature creates multiple fragments with identical name
Sometimes when an ‘unlinked’ fragment is converted to an ‘anonymous’ one, it’s given an ‘Anonymous F-’ number which is already assigned to an existing anonymous fragment. Both fragments continue to exist with the identical number.

**To Reproduce**
1. Open any fragment for editing, and click the button to edit the ‘commentary’ box.
2. Type ‘@F198’ to create a link to fragment ‘Anonymous F198’.
3. In the menu that pops up, ‘Anonymous F198’ will appear twice.
4. Create two links, one pointing to each ‘Anonymous F198’.
5. Click on the links and see that there are two separate fragments called ‘Anonymous F198’.

**Expected behavior**
When an unlinked fragment is converted to an anonymous fragment, it should be given a new and not yet assigned anonymous F- number.

**Additional context**
Note: This has happened to me a few times while trying to convert relatively old unlinked and anonymous fragments back and forth. I have tried to reproduce the problem with new fragments created for the purpose, but for some reason the problem doesn’t seem to occur with freshly minted fragments.

tcouch:
closes #309 
Anonymous fragments needed to be reindexed after an unlinked fragment is made anonymous. Reindexing removes any collisions caused by two anon fragments having the same order value.

### DIFF
--- a/src/.pre-commit-config.yaml
+++ b/src/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: check-yaml
 
   - repo: https://github.com/psf/black
-    rev: 21.8b0
+    rev: 22.3.0
     hooks:
       - id: black
 
--- a/src/.pre-commit-config.yaml
+++ b/src/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: check-yaml
 
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: stable
     hooks:
       - id: black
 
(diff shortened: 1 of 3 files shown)


src/.pre-commit-config.yaml
```diff
@@ -11,7 +11,7 @@ repos:
       - id: check-yaml
 
   - repo: https://github.com/psf/black
-    rev: 21.8b0
+    rev: 22.3.0
```
raquelalegre: I have noticed [they have a tag named "stable"](https://github.com/psf/black/tags) that might always point to the latest version in case that helps not needing to update the version here when it changes. The problem is it might introduce problems with a new version change and we won't notice so easily the problem might come from here.

raquelalegre (approved):
Looks good! 
Still unsure about the change of black version.

src/.pre-commit-config.yaml
tcouch: Yep, changed it to stable and it still works so I'll go with that. Thanks!

PR #330, opened 2022-06-21
Fix/missing_frags_on_work_detail

### LINKED ISSUE #325: Missing fragments on citing work pages
**Describe the bug**
While working on the entry for Gavius Bassus, we found that some fragments do not appear when ‘citing work’ is selected; removing and adding identifications causes fragments to be renumbered but does not seem to solve the problem.

tcouch:
closes #325 missing fragments on citing work pages.

The problem was in the make_grouped_dict function which was overwriting lists of definite fragments if possible fragments existed.

### DIFF
--- a/src/.pre-commit-config.yaml
+++ b/src/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: check-yaml
 
   - repo: https://github.com/psf/black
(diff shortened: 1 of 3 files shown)


raquelalegre (approved):
Looks good!

PR #332, opened 2022-06-22
Display testimonia at the beginning of the ordered material section.

### LINKED ISSUE #324: Reorder testimonia on antiquarian details page
Testimonia that relate only to an antiquarian are currently grouped under ‘unknown work’, these should come immediately under ‘Ordered Material’ and before the first of the titles (e.g. Lucius Accius T3 should actually appear above the title ‘Didascalica’ and before ‘Ordered Material’; it should thus be T1 [=Lucius Accius T1])

raquelalegre:
Don't display as unknown work.
Fixes #324


raquelalegre:
I just realised testimonia that is not assigned to any book were appearing too many times. The `div` was mistakenly inside the for loop that goes through the list of works.

This still needs testimonia IDs to be reworked so the first testimonium showing in the page of an antiquarian is T1, but I'll leave that to @tcouch. Tom, feel free to modify to fix the testimonium ID problem, or merge, as you prefer!

tcouch (approved):

PR #342, opened 2022-08-09
Merge latest changes into production

### LINKED ISSUE #317: Search within specific fragment content fields
We'd need for Anon Fragment,  Fragment and Testimonium the following 3 subcategories:
* Original Text
* Translation
* Commentary

In the future this could be done in a more sophisticated way with a dynamic dropdown that's hidden and only shows when one of the models mentioned is chosen. See #318.

### LINKED ISSUE #326: Allow users to order anonymous fragments within topics
Splitting this out from #290 as these requests can be tackled independently.

## Description
On the /anonymous/list page, anonymous fragments are currently displayed under topic headings and the order in which they're displayed within those topics is determined by the fragment's order value. Users would like to be able to change the fragment's order within topics by dragging them up and down (or clicking up/down buttons) within the topic list (c.f. /topic/list/ page).

## Current functionality
* Anonymous fragments have an order field (OrderableModel) which is a positive integer initially assigned as the last position in the list of anon fragments when the object is first saved; e.g. if there are 100 anon fragments, a new fragment would be given the order value 100 (order starts counting from 0). Order is then used to generate the display name for the fragment, which in this case would be F101 (the get_display_name function adds 1 to the order).
* Anonymous fragments have a many-to-many relationship with Topics (they can belong to many topics, and topics can relate to many fragments). E.g. if an AnonymousFragment is assigned to three topics, they will appear three times in the anonymous section, once under each topic.
* Topics also have an order, which can be changed by users on the /topic/list page.
* Every time an anonymous fragment is added to, or removed from a topic, or the topic order changes, the order values of all anonymous fragments are recalculated. To do this: a `reindex_anonymous_fragments` function creates a queryset containing all AnonymousFragments ordered by topic order, then by their own order value; duplicates are removed; then order values are updated for each fragment according to its position in the queryset.
* Anonymous fragments which are not assigned to any topic are listed in the “Uncategorised” section and are implicitly treated as belonging to a topic with the highest order value.

## Proposed solution
Luckily @peterbispham seems to have anticipated this and added an order field to AnonymousTopicLink which we can use to record the order with respect to Topic for Anonymous Fragments. A quick survey of the database confirms this field is not currently used.

Once we have an order wrt topic for each AnonymousTopicLink, we can change the reindex_anonymous_fragments function to order AnonymousFragments by Topic order, then by AnonymousTopicLink order to determine their overall order.

tcouch:
closes #326
closes #317 

# Search within specific fragment content fields
Adds the ability to search within original text content, translations and commentaries separately, rather than all at once. To achieve this, new search methods have been added, and the content type drop-down list on the search page now splits fragments, testimonia, anonymous fragments, and apposita into groups. See https://www.ucl.ac.uk/history/frrant-preprod/search/

# Allow users to order anonymous fragments within topics
The following changes have been made to the [anonymous fragment list page](https://www.ucl.ac.uk/history/frrant-preprod/anonymous/list/):
* Pagination has been removed
* Instead, select a topic from the drop-down menu to view anonymous fragments related to that topic
* Use up/down buttons, or drag fragments to reorder them within the topic
* Overall anonymous fragment order (and hence their name) is now determined first by topic order, and then by order within topic.

Example:
Anonymous fragments A and B  both belong to the Monarchy topic and the Law topic. Monarchy is ordered first, followed by Law. Under Monarchy A is ordered first, and then B. Under Law B is ordered first, and then A. In this situation fragment A would be named "Anonymous F1" and B would be named "Anonymous F2" because their respective order under the first topic takes precedence over their order in the second.


tcouch:
Merging this into production now since everyone had a look yesterday and no issues were raised.

PR #345, opened 2022-08-25
Feature/multiple ref inputs

### LINKED ISSUE #321: Allow multiple reference inputs when creating fragment
Would it be possible to add additional boxes to the ‘Reference’ portion of the create fragment / anonymous fragment / testimonium pages?
Something like the following:
image

Where Reference 1 / 2 is an alphanumeric input (e.g. 2.3.4, p. 123), Reference 1 Source / 2 Source is an alphabetical input (e.g. Smith, Lindsay)
Reference order would be used as currently but be based on the Reference 1 entry.
The additional boxes could work like the concordance section (e.g. with an ‘add additional reference’ button).
This would allow us to refer to texts in multiple ways according to different editions.
As with many of our requests, it would be useful to know how much work would be involved in implementing this, and we could then assign an appropriate priority level.

acholyn:
closes #321 
also addresses issues with tests and includes new ones for references


src/rard/research/models/original_text.py
```diff
@@ -19,28 +20,47 @@ def related_lock_object(self):
...
         ordering = ("citing_work", "reference_order")
 
+    @property
+    def reference(self):
+        references = Reference.objects.filter(original_text=self)
+        if references:
+            # only return the number values as the editor isn't critical
+            return references.first().reference_position
+        else:
+            return ""
```
tcouch: We can get rid of this now right? Or are there templates using it somewhere?

src/rard/research/tests/models/test_original_text.py
```diff
@@ -203,3 +204,29 @@ def test_no_edition_in_display(self):
...
+            citing_work=self.citing_work,
+        )
+
+        reference = "2.6-9"
+        self.reference = Reference.objects.create(
+            editor="superfluous",
+            reference_position=reference,
+            original_text=self.original_text,
+        )
+
```
tcouch, line 225: I think we need a test_reference_list_property test here. So, make sure that when there's only one reference you just get the reference position, and when there's two you get "editor1 1.2.3 | editor2 4.5.6"

src/rard/templates/research/partials/section/original_text_details.html
```diff
@@ -18,18 +18,28 @@
...
-
+    {% comment %} <dd class="col-sm-9">{{ original_text.remove_reference_order_padding }}</dd>
+    <dd class="col-sm-9">
+    {% for i in original_text.references.all %}
+        <p>
+            {{i.order}}
+        </p>
+    {% endfor %}
+    </dd>
+{% endcomment %}
```
tcouch: We can just remove this right?

tcouch (changes requested):
I've made a few changes including:
* splitting the migration into 2 because it was breaking when I tried to roll it back. Setting a default to the removed fields before deleting them fixed that
* various test fixes
* the reference order field was missing from the create template
* linting
In general can you run pre-commit and make sure all tests are passing locally before creating a PR?

src/rard/research/models/original_text.py
acholyn: I think it's good to go; the forms shouldn't be affected

src/rard/research/models/original_text.py
```diff
@@ -19,28 +20,38 @@ def related_lock_object(self):
...
+        """Returns a list of references with editor and ref position for each
+        unless there's only one, in which case only the ref position is shown"""
+        references = Reference.objects.filter(original_text=self)
+        if references.count() > 1:
+            return [
+                f"{reference.editor} {reference.reference_position} |"
+                for reference in references
+            ]
+        else:
+            return [references.first().reference_position]
```
tcouch: Please don't kill me, but this property could just return a string rather than a list right?
```suggestion
    def reference_list(self):
        """Returns a string which lists the editor and ref position for each reference
        unless there's only one, in which case only the ref position is shown"""
        references = Reference.objects.filter(original_text=self)
        if references.count() > 1:
            return " | ".join([
                f"{reference.editor} {reference.reference_position}"
                for reference in references
            ])
        else:
            return references.first().reference_position
```

src/rard/templates/research/partials/section/original_text_details.html
```diff
@@ -18,18 +18,17 @@
         {% endif %}
     </dd>
 
-    <dt class="col-sm-3">{% trans 'Reference' %}</dt>
-    <dd class="col-sm-9">{{ original_text.reference|default:'-' }}</dd>
+    <dt class="col-sm-3">{% trans 'References' %}</dt>
+    <dd class="col-sm-9">
+    {% for i in original_text.reference_list %}{{i}}{% endfor %}
```
tcouch: If reference_list is just a string:
```suggestion
    {{original_text.reference_list}}
```

src/rard/research/tests/models/test_original_text.py
```diff
@@ -203,3 +204,44 @@ def test_no_edition_in_display(self):
...
+    def test_reference_deletion(self):
+        self.reference.delete()
+        self.original_text.refresh_from_db()
+        self.assertEqual(self.original_text.references.count(), 0)
+
+    def test_reference_list_property(self):
+        self.assertEqual(self.original_text.references.count(), 1)
+        self.assertEqual(
+            self.original_text.reference_list[0], self.reference.reference_position
+        )
```
tcouch:
```suggestion
        self.assertEqual(
            self.original_text.reference_list, self.reference.reference_position
        )
```

src/rard/research/tests/models/test_original_text.py
```diff
@@ -203,3 +204,44 @@ def test_no_edition_in_display(self):
...
+        self.reference2 = Reference.objects.create(
+            editor="superfluous",
+            reference_position="3.9.9",
+            original_text=self.original_text,
+        )
+        references_list = ["superfluous 2.6-9 |", "superfluous 3.9.9 |"]
+        self.assertEqual(self.original_text.references.count(), 2)
+        self.assertIn("superfluous", self.original_text.reference_list[0])
+        self.assertEqual(self.original_text.reference_list, references_list)
```
tcouch:
```suggestion
        references_list = "superfluous 2.6-9 | superfluous 3.9.9 "
        self.assertEqual(self.original_text.references.count(), 2)
        self.assertEqual(self.original_text.reference_list, references_list)
```

tcouch (commented):
Would this work? Have I forgotten any other cases where we need reference_list to actually be a list?

src/rard/research/models/original_text.py
acholyn: yes that would work

PR #349, opened 2022-09-06
Feature/dynamiclinks

### LINKED ISSUE #202: Dynamic links filtering
We can do something like this in order to make it easier to mention fragments/authors/etc once the DB grows more: https://www.w3schools.com/howto/howto_js_filter_dropdown.asp

acholyn:
Closes #202, includes tests


tcouch:
@acholyn I think it's be good if we could do the same thing for work searches as we've done for bibliography items; i.e. create an annotation that combines antiquarian name and work name and search that.

acholyn:
The annotation method could be extracted into its own function but I've opted to just add the work search as another function

src/rard/research/views/mention.py
```diff
@@ -26,102 +27,145 @@ class MentionSearchView(LoginRequiredMixin, View):
...
-        if len(k) == 0:
-            return keywords[1:]
-        keywords[0] = k
-        return keywords
+    def work_search(cls, keywords):
+        qs = Work.objects.annotate(
+            author_title=Concat(
+                F("authors"), Value(" "), F("name"), output_field=CharField()
+            )
+        )
```
tcouch: I noticed authors just uses the antiquarian's order name (e.g. Accius for Lucius Accius). That means if you do a work search for `@wk:lucius` it won't show any of Accius's works. We can get the whole name by copying the pattern used by the WorkManager which uses StringAgg to add the authors annotation, but use antiquarian__name instead.

```suggestion
        qs = Work.objects.annotate(
            author_title=Concat(
                StringAgg("antiquarian__name", delimiter=","),
                Value(" "),
                F("name"),
                output_field=CharField()
            )
        )
```

src/rard/research/views/mention.py
```diff
@@ -1,9 +1,10 @@
 # from django.contrib.postgres.search import SearchQuery, SearchRank, \
 #     SearchVector
-from itertools import chain
 
+from django.apps import apps
 from django.contrib.auth.mixins import LoginRequiredMixin
-from django.db.models import Q
+from django.db.models import CharField, F, Q, Value
```
tcouch, line 7: Import StringAgg (see suggestion below)
```suggestion
from django.db.models import CharField, F, Q, Value
from django.contrib.postgres.aggregates import StringAgg
```

src/rard/research/tests/views/test_mentions.py
```diff
@@ -64,131 +116,263 @@ def test_default_queryset(self):
...
-            parent=a2,
+        self.assertEqual(len(list(view.get_queryset())), 0)
+
+        # string as search
+        view.request = self.request(
+            data={
+                "q": "wk:ro",
+            }
+        )
+        self.assertEqual(len(list(view.get_queryset())), 3)
```
tcouch: I don't think the number search is necessary here, but what's missing is a test to see if we're matching work authors' names. If you add works to antiquarians in the setup function we can test that here; e.g. "and" should match antiquarian "Andrew" and work "andropov"

src/rard/research/tests/views/test_mentions.py
```diff
@@ -64,131 +116,263 @@ def test_default_queryset(self):
...
+        )
+
+        # topics
+        self.assertEqual(len(list(view.basic_search(view, "tp", ""))), 3)
+        self.assertEqual(
+            list(view.basic_search(view, "tp", "s")),
+            [self.pictures, self.coups, self.estates],
+        )
+
+    def test_bibliography_search(self):
```
tcouch, line 149: I can't see any examples here where you try to match the contents of bibliography titles. It'd be good to test for something that matches across both fields on different objects; e.g. "two" matching bi 2 and bi3 if the title for bi3 was "peppers two apples"

tcouch (changes requested):

src/rard/research/views/mention.py
acholyn: oh nice, thanks!

tcouch:
Great. I think this looks good now.

Issue #351, opened 2022-09-06
Original texts can be created with no reference


tcouch:
When creating a new fragment, the form can be submitted with no references. The fragment is created successfully, but an error is raised when rendering the detail page because the reference_list property cannot deal with an original text having no references.

The solution would be to either:
* require all original texts to have at least one reference
* fix the reference_list property to allow for no references

tcouch:
@rmamarshall do you think it would be preferable to allow original texts to be created without any references, or should be put controls in place to prevent that happening?

rmamarshall:
@tcouch Every original text needs at least one reference - a block to prevent the creation of original texts without an entry in a reference field would be a good idea.

PR #352, opened 2022-09-06
Merge latest changes into production

tcouch:
Change summary:
* Allow multiple reference inputs when creating fragments
* Improved \@mention search
* Wildcard search

## Allow multiple reference inputs when creating fragments
Instead of being a property of original texts, references are now an object in their own right (with editor and reference fields). With this change original texts can have one or more references instead of just one. Users can add multiple reference when creating the fragment/original text by clicking "Add another reference", and this option is also available when editing an existing original text.

The newly created reference model has three fields:
* original_text (the owner) (not viewable on the form as it is implicit)
* reference_position (used to refer to the position in the text e.g. 2.6-8)
* editor (optional name of the editor)

For sorting purposes, original texts have kept their reference position property, which can be seen when editing the item. However, this property is not displayed on any detail or list pages.

Where an original text only has one reference, its reference position is displayed on the fragment detail page like this:
> 1.2.3

Where there are multiple references, editor name and reference position is displayed for each reference like this:
> AuthorA 1.2.3 | AuthorB 4.5.6

### Data migration
For each existing original_text, its reference property (a character field) is converted into a new reference object, owned by the original text. The reference property becomes the reference_position property on the newly created reference, and the editor field is blank.

## Improved \@mention search
Backend:
* All \@mention searches begin with a two letter short code indicating the target model type (see table below)
* Remove blanket search - choose search method based on initial two-letter code
* Instead of treating everything after the @ symbol as one keyword to match exactly, allow users to enter multiple keywords separated by colon character; e.g. @<wbr>aq:marcus:varro
* Move sorting to search methods so we can tune it to the search method

| Model              | Short code | Sort by              | Example    |
|--------------------|------------|----------------------|------------|
| Antiquarian        | @<wbr>aq        | order_name desc      | @<wbr>aq:varro  |
| Bibliography       | @<wbr>bi        | author_surnames, year | @<wbr>bi:music:jones  |
| Topic              | @<wbr>tp        | topic order          | @<wbr>tp:war    |
| Work               | @<wbr>wk        | name                 | @<wbr>wk:musica:varro |
| Fragment           | @<wbr>fr        | pk  | @<wbr>fr:varro  |
| Anonymous Fragment | @<wbr>af        | order                | @<wbr>af:12     |
| Testimonium        | @<wbr>tt        | antiquarian_testimoniumlinks__antiquarian__name                  | @<wbr>tt:varro  |
| Unlinked Fragment  | @<wbr>uf        | pk                   | @<wbr>uf:123    |

## Wildcard Search
The search page now supports searching with wildcard characters:

* \* represents zero or more characters; e.g. *ara* matches paragraph, ara, and chiara
* ? represents exactly one character; e.g. ?ara* matches paragraph, but not ara or chiara
* Wildcards can be used within quotes as follows: "Veni * Vi?i" matches "Veni vidi vici"
* Searches for keywords without wildcards now only return exact matches; e.g. ara now only matches ara and not paragraph


tcouch:
One potentially controversial decision is how we're diplaying fragments' summary information in lists. In the image below, the citing work information includes all references including editor information in the same format as displayed on detail pages if there are multiple references. This is fairly trivial to change.

image

tcouch:
This change can now be tested on the development server here: https://www.ucl.ac.uk/history/frrant-preprod/fragment/create/

tcouch:
#351 needs to be addressed before we merge this into production (original texts can currently be created with no references).

tcouch:
Ok, #351 is fixed now so all original texts must have at least one reference.

tcouch:
Just discovered #354 which prevents users from adding a new original text to an existing fragment. Needs fixing before we can merge.

tcouch:
#354 is fixed now

Issue #354, opened 2022-09-13
Cannot add original text to fragment


tcouch:
Following the update to make it possible to add multiple references to a fragment, we get an error when trying to add an original text to an existing fragment.

`bootstrap4.exceptions.BootstrapError: Parameter "field" should contain a valid Django BoundField.`

PR #355, opened 2022-09-14
Fix/cannot-create-original-text

tcouch:
Fix for #354.

* Removed mention of original_text.reference field from originaltext_form.html
* Added reference formset to create_form context
* Include reference_form.html in originaltext_form.html
* Fix original_text tests
* Added reference formset tests


acholyn (approved):
looks nice and familiar

PR #367, opened 2022-10-24
remove dump file after loading data

### LINKED ISSUE #346: Dumps need updating
Currently the dumps are including sessions.session information, which isn't needed in the json file - update documentation to use something more sophisticated - exclude auth and other possible troublemakers (sites.site), alternatively it might be simpler to only include the research data

It would also be good to update the `loaddata` script to remove `dump.json` when it's finished

acholyn:
closes #346 
has now been tested

### DIFF
--- a/src/loaddata.sh
+++ b/src/loaddata.sh
@@ -13,7 +13,7 @@ then
 echo ${environment} environment
 docker cp $1 ${container}:/app/dump.json
 docker exec -it ${container} /bin/bash -c ". /entrypoint && LOADING=true ./manage.py loaddata /app/dump.json"
-docker exec ${container} /bin/bash -c "rm -rf /app/dump.json"
+docker exec ${container} /bin/bash -c "rm /app/dump.json"
 exit 0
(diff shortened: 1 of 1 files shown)


Devaraj-G:
Apologies, I am not confident/lack knowledge to review this.

src/loaddata.sh
```diff
@@ -13,7 +13,9 @@ then
 echo ${environment} environment
 docker cp $1 ${container}:/app/dump.json
 docker exec -it ${container} /bin/bash -c ". /entrypoint && LOADING=true ./manage.py loaddata /app/dump.json"
+docker exec ${container} /bin/bash -c "rm -rf /app/dump.json"
```
tcouch:
```suggestion
docker exec ${container} /bin/bash -c "rm /app/dump.json"
```
-rf shouldn't be necessary since we're deleting a file rather than a directory and there's no need to force it.

tcouch (approved):

PR #373, opened 2023-01-13
Feature/323-update-bibliography-links-from-mentions

tcouch:
The research team would like items to be added to Antiquarian.bibliography_items automatically when referenced via an @mention in the introduction.

### DIFF
--- a/src/rard/research/models/text_object_field.py
+++ b/src/rard/research/models/text_object_field.py
@@ -41,12 +41,8 @@ class TextObjectField(HistoryModelMixin, BaseModel):
         return None
 
     def save(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


tcouch (commented):
We need guidance on how to deal with mentions that are later removed from the introduction. Either:

- A. The introduction content alone determines which bibliography items should be linked to the antiquarian. In which case we can adapt this to re-scan the introduction content each time it is saved and update the whole Antiquarian.bibliography_items collection.
- B. Bibliography items can also be added independently of @mentions, but those added via @mentions should be removed if the mention is deleted. In this case we probably need a through table on the m2m relationship to track the source of the link.

PR #375, opened 2023-01-30
Fix/fragment ordering

acholyn:


src/rard/research/views/antiquarian.py
```diff
@@ -77,10 +78,10 @@ def post(self, *args, **kwargs):
...
                 model_class = model_classes[object_type]
                 link = model_class.objects.get(pk=link_pk)
-                if "up_by_work" in self.request.POST:
-                    link.move_to_by_work(link.work_order - 1)
-                elif "down_by_work" in self.request.POST:
-                    link.move_to_by_work(link.work_order + 1)
+                if "up_by_book" in self.request.POST:
+                    link.move_to_by_book(link.work_order - 1)
+                elif "down_by_book" in self.request.POST:
+                    link.move_to_by_book(link.work_order + 1)
```
tcouch: It's order_in_book +/- 1 we want to pass as the new position here I think:
```suggestion
                if "up_by_book" in self.request.POST:
                    link.move_to_by_book(link.order_in_book - 1)
                elif "down_by_book" in self.request.POST:
                    link.move_to_by_book(link.order_in_book + 1)
```

tcouch:
A few notes on the current state of the Work detail page:
* Unknown book shouldn't have the up/down buttons since it should always be at the bottom (you get an error when clicking these at the moment)
* Reordering books needs to trigger reindexing wrt work so the Fragment names get update
* If you look at `/work/22/` for example, the Apposita sub headings appear the same size as the book headings.
* On the Antiquarian detail pages, it looks like Anonymous Fragment links don't have a possible/definite indicator

tcouch (commented):
This seems to have broken the content replacement when you click an up/down button. It all disappears.

tcouch (commented):
Now books with no content don't appear when the page is first loaded, but do show up after the ajax request.

tcouch:
### Adding unknown works to all Antiquarians and unknown books to all works

1. Add a new boolean field to Work and Book models. This will be false by default (so as not to disrupt any existing code), but will be set to True for unknown Works and Books. Should enforce only one Work per Antiquarian can have unknown=True, likewise for Books in Works.
2. Edit the migration file to loop through every Antiquarian and add an unknown Work (name="Unknown Work", unknown=True), then loop through every Work and add an unknown book (name="Unknown Book", unknown=True). Then loop through every FragmentLink, TestimoniumLink, and AppositumLink and:
  * Where work is currently None, set it to the link's Antiquarian's unknown work
  * Where book is currently None, set it to the link's Work's unknown book
3. Add a Antiquarian.unknown_work property to make it easier to retrieve it's unknown work. Do the same for Work.unknown_work
4. When an Antiquarian is created, automatically create an unknown work
5. When a Work is created, automatically create an unknown book
6. Change default ordering so that unknown work/book is always at the end
7. Users should not be able to move unknown works/books up/down
8. On detail page do not show unknown work/book if there's no content to display

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,108 @@
...
+        Work = apps.get_model("research", "Work")
+        works = Work.objects.all()
+        for work in works:
+            work.add_book(name="Unknown Book", unknown=True)
+
+    def assign_links_to_unknown(apps, schema_editor):
+        worklinks = ["FragmentLink", "TestimoniumLink", "AppositumFragmentLink"]
+
+        for link_type in worklinks:
+            links = link_type.object.all()
```
tcouch: You need to get the actual Class in order to use `.objects.all()`
```suggestion
            link_class = apps.get("research", link_type)
            links = link_class.object.all()
```

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,108 @@
...
+            for link in links:
+                ant = link.antiquarian
+                if link.work == None:
+                    link.work = ant.works.get(unknown=True)
+                work = link.work
+                if link.book == None:
+                    link.book = work.books.get(unknown=True)
```
tcouch: I think this has to be `work.book_set.` because the book model doesn't specify a related name on the foreign key

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def give_books_order(apps, schema_editor):
+        Work = apps.get_model("research", "Work")
+        for work in Work.objects.all():
+            books = work.book_set.all().order_by("number")
```
tcouch: Now that we have the unknown boolean field, I think we want to and this to order_by such that unknown=True, appears at the end regardless of "number".

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,108 @@
...
+
+        def order_in_book_for_link_class(link_classname):
+            link_class = apps.get_model("research", link_classname)
+            for work in Work.objects.all():
+                related_links = link_class.objects.filter(work=work).order_by(
+                    "book__order", "work_order"
+                )
+                for count, link in enumerate(related_links):
+                    link.order_in_book = count
+                    link.save()
```
tcouch: I think this isn't quite right. We want order_in_book to be 0 to n-1 for the n links associated with each Book, but this would set order_in_book to 0 to m-1 for the m links associated with each Work. So, we want to loop through each Book, get related links ordered by work_order, then loop through the queryset and assign order_in_book.

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,196 @@
...
+
+        def order_in_book_for_link_class(link_classname):
+            link_class = apps.get_model("research", link_classname)
+            for work in Work.objects.all():
+                related_links = link_class.objects.filter(work=work).order_by(
+                    "work_order"
+                )
+                for count, link in enumerate(related_links):
+                    link.order_in_book = count
+                    link.save()
```
tcouch: I think we need to be iterating through Books rather than Works here. As written, related_links will pick up all links belonging to the work, but for order_in_book, that's not what we want: For each Book we want to count from 1-N where N is the number of links belonging it.

**Note:** Still using work_order to order related_links though because we're starting off by deriving order in book from order in work.
 
```suggestion
        def order_in_book_for_link_class(link_classname):
            link_class = apps.get_model("research", link_classname)
            for book in Book.objects.all():
                related_links = link_class.objects.filter(book=book).order_by(
                    "work_order"
                )
                for count, link in enumerate(related_links):
                    link.order_in_book = count
                    link.save()
```

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,196 @@
...
+    def give_books_order(apps, schema_editor):
+        Work = apps.get_model("research", "Work")
+        for work in Work.objects.all():
+            books = work.book_set.all().order_by("unknown", "number")
+            for count, book in enumerate(books):
+                book.order = count
+                book.save()
+
+    def give_links_order_in_book(apps, schema_editor):
+        worklinks = ["FragmentLink", "TestimoniumLink", "AppositumFragmentLink"]
```
tcouch:
```suggestion
    def give_links_order_in_book(apps, schema_editor):
        """As a starting point, order_in_book, will inherit its order from work_order;
        i.e. if link A precedes link B in work_order, it will precede link B in order_in_book
        if both belong to the same Book."""
        worklinks = ["FragmentLink", "TestimoniumLink", "AppositumFragmentLink"]
```

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,196 @@
...
+                )
+                for count, link in enumerate(related_links):
+                    link.order_in_book = count
+                    link.save()
+
+        for link in worklinks:
+            order_in_book_for_link_class(link)
+
+    def reindex_work_by_book(apps, schema_editor):
+        """Update order of links with respect to work, taking into account book__order and order_in_book"""
```
tcouch:
```suggestion
        """A link's order represents its order wrt the Antiquarian. This is derived first of all 
            from the order of the link's work wrt the Antiquarian, then from the link's order with 
            respect to the work (work_order). Work order is derived in turn from the order of
            the link's book with respect to the Work, and then the link's order within that book.
            
            We now need to loop through each Work and recalculate work_order from the new
            order_in_book and book__order properties that have been added. Then we can
            loop through each Antiquarian and recalculate each related link's order based on 
            work__worklink__order and work_order."""
```

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,196 @@
...
+
+                for count, link in enumerate(to_reorder):
+                    if link.book.work.unknown == True:
+                        link.work_order = link.order_in_book
+                        link.save()
+                    else:
+                        if link.work_order != count:
+                            link.work_order = count
+                            link.save()
+                    reindex_fragment_and_testimonium_links(link.antiquarian)
```
tcouch:
```suggestion
        Antiquarian = apps.get_model("research", "Antiquarian")

        def reindex_links_to_work(work):
            for class_name in worklinks:
                link_class = apps.get_model("research", class_name)

                # Unknown book's links should always be last
                # Then order by book order, then order in book
                to_reorder = (
                    link_class.objects.filter(work=work)
                    .order_by("book__unknown", "book__order", "order_in_book")
                    .distinct()
                )

                for count, link in enumerate(to_reorder):
                    if link.work_order != count:
                        link.work_order = count
                        link.save()
        
        def reindex_links_to_antiquarian(antiquarian):
            for class_name in worklinks:
                link_class = apps.get_model("research", class_name)
                
                to_reorder = (
                    link_class.objects.filter(antiquarian=antiquarian)
                    .order_by("work__unknown", "work__worklink__order","work_order")
                    .ditinct()
                )
                
                for count, link in enumerate(to_reorder):
                    if link.order != count:
                        link.order = count
                        link.save()

        for work in Work.objects.all():
            reindex_links_to_work(work)
            
        for ant in Antiquarian.objects.all():
            reindex_links_to_antiquarian(ant)
        
```

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,196 @@
...
+                    "work__worklink__order",
+                    "work_order",
+                    "book__order",
+                    "order_in_book",
+                )
+                .distinct(),
+                # We want testimonium links without works to be ordered first
+                # but for some reason adding "-work__isnull" to the order_by
+                # breaks the rest of the ordering. Using itertools chain as
+                # a workaround does the job.
```
tcouch: No longer relevant because there are no links without works

src/rard/research/migrations/0061_order_in_book.py
```diff
@@ -0,0 +1,196 @@
...
+
+                to_reorder = (
+                    link.objects.filter(work=work)
+                    .order_by("book__order", "order_in_book")
+                    .distinct()
+                )
+
+                for count, link in enumerate(to_reorder):
+                    if link.book.work.unknown == True:
+                        link.work_order = link.order_in_book
```
tcouch: I'm not sure what this is supposed to achieve, why not just include book__unknown in the order_by above?

src/rard/research/migrations/0061_order_in_book.py
tcouch: Calling reindex_fragment_and_testimonium_links within this loop will call it once for each link in the work. Best to do reindexing link order at the end once work_order is sorted.

tcouch (changes requested):
Just gone through the migration first, but I've noticed a few things that could do with looking at/fixing

src/rard/research/migrations/0061_order_in_book.py
acholyn: I think this was the method used on the base model and we copied it over for an error? iirc

src/rard/research/forms.py
```diff
@@ -616,6 +616,22 @@ class Meta:
...
+# FragmentInlineUpdateFormset = inlineformset_factory(
+#     Book,
+#     Work,
+#     form=InlineFragmentForm,
+#     fields=("book", "work"),
+#     labels={"book": "Book", "work": "Work"},
+#     can_delete=False,
+# )
+
+
```
tcouch: Can this go?

src/rard/research/models/work.py
```diff
@@ -38,9 +42,14 @@ class Meta:
...
     number_of_books = models.CharField(max_length=128, blank=True)
 
-    # antiquarians = models.ManyToManyField('Antiquarian', blank=True,
-    #     through='WorkLink'
-    # )
+    unknown = models.BooleanField(default=False)
+
+    def book_set(self):
+        return super().book_set().order_by("unknown", "order", "number")
+
```
tcouch: Not needed

src/rard/templates/base.html
```diff
@@ -173,6 +181,7 @@
           {% trans "Roman Republican Antiquarians Research Database" %}
           {% if not PRODUCTION_INSTANCE %}
             (NON PRODUCTION VERSION)
+            <div class="non-prod-column"></div>
```
tcouch, line 184: Nice

src/rard/templates/research/base.html
```diff
@@ -212,6 +206,12 @@
         {% trans 'Anonymous Fragments' %}
       </a>
       {% endif %}
+      {% if perms.research.view_fragment %}
+      <a class='nav-link {% if "unlinked_fragment" in request.resolver_match.namespaces %}active font-weight-bold{% endif %}'
+        href='{% url "unlinked_fragment:list" %}'>
+        {% trans 'Unlinked Fragments' %}
+      </a>
+      {% endif %}
```
tcouch, line 214: Good addition. Should really be in a separate pull request.

src/rard/templates/research/partials/render_inline_fragment_form.html
```diff
@@ -0,0 +1,11 @@
+{% load crispy_forms_tags %}
+{% if can_edit %}
+
+{% comment %} selects contain the books and works of the antiquarian and will work by displaying the work first and then display the books in that work {% endcomment %}
+<div>
+<select>
+</select>
+
+</div>
+
```
tcouch, line 10: If this is work in progress, best not to submit it in this PR.

src/rard/research/tests/models/test_antiquarian.py
```diff
@@ -12,7 +12,9 @@
 class TestAntiquarian(TestCase):
     def test_creation(self):
         data = {"name": "John Smith", "re_code": "smitre001"}
-        a = Antiquarian.objects.create(**data)
+        a = Antiquarian.objects.create(
+            **data
+        )  # fails here; rard.research.models.work.Book.work.RelatedObjectDoesNotExist: Book has no work.
```
tcouch: Is this comment meant to be here?

src/rard/research/tests/models/test_link_scheme.py
```diff
@@ -756,6 +773,9 @@ def test_testimonium_queryset_methods(self):
         self.assertEqual(0, self.testimonium.definite_antiquarian_links().count())
         self.assertEqual(1, self.testimonium.possible_antiquarian_links().count())
 
+    @pytest.mark.skip(
```
tcouch: Probably best to have this working so that we remember to investigate if work on changes to definite status break it.

src/rard/research/tests/models/test_link_scheme.py
```diff
@@ -316,6 +316,7 @@ def test_link_antiquarian_orders_sequentially(self):
         ):
             self.assertEqual(count, link.order)
 
+    @pytest.mark.skip("the opposite of this would be true now")
```
tcouch: Testimonia are a special case where those belonging to unknown work should be ordered first. Antiquarian reindex method needs updating to fix this.

tcouch (commented):
Some fixes for these are included in the following PR.

src/rard/research/forms.py
acholyn: I believe that's the basis of the inline form for updating the links, so I can do since we'll be covering it in the other PR

src/rard/research/tests/models/test_link_scheme.py
acholyn: ah...I'll need to fix that in the new feature as unknown is at the end I believe

src/rard/research/tests/models/test_link_scheme.py
acholyn: I've fixed it in the new feature. I can work on fixing it for this PR but will need to rebuild the db and everything

tcouch (approved):

PR #384, opened 2023-04-24
Fix/fragment ordering tc2

tcouch:
Changes include:
* Revert to putting Testimonia without works at top of Antiquarian page
* Remove unknown work from names
* Fix reordering testimonia and apposita on work detail pages
* Fix reindexing following book move
* Some tests reinstated
* Handle book deletion

### DIFF
--- a/src/rard/research/models/fragment.py
+++ b/src/rard/research/models/fragment.py
@@ -55,7 +55,6 @@ post_delete.connect(handle_deleted_topic_link, sender=TopicLink)
 
 
 class Fragment(HistoryModelMixin, HistoricalBaseModel, DatedModel):
-
     history = HistoricalRecords(
         excluded_fields=[
             "topics",
@@ -75,6 +74,7 @@ class Fragment(HistoryModelMixin, HistoricalBaseModel, DatedModel):
         return (
             self.antiquarian_fragmentlinks.filter(
                 definite=True,
+                work__unknown=False,
                 work__isnull=False,
             )
             .order_by("work", "-book")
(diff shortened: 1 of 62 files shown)


tcouch:
@acholyn Sorry I just realised having running pre-commit and updated loads of files makes it really hard to see what's changed. Mostly models.base, models.antiquarians, models.work, and the migration file.

src/rard/research/models/fragment.py
```diff
@@ -75,6 +74,7 @@ def definite_work_and_book_links(self):
         return (
```
acholyn, line 74: Does it matter that we overhaul this in the next PR? (The definite/possible grouping)

src/rard/research/tests/models/test_link_scheme.py
```diff
@@ -773,9 +773,6 @@ def test_testimonium_queryset_methods(self):
         self.assertEqual(0, self.testimonium.definite_antiquarian_links().count())
         self.assertEqual(1, self.testimonium.possible_antiquarian_links().count())
 
-    @pytest.mark.skip(
-        "This can be revisited when definite status is broken down into ant/wk/bk"
-    )
     def test_add_antiquarian_fragment_ignores_work(self):
```
acholyn: does this test pass for you then?

acholyn (approved):
Looks good, thank you!

src/rard/research/models/fragment.py
tcouch: I think it's best to keep interdependencies between PRs to a minimum, so we know this version "works".

src/rard/research/tests/models/test_link_scheme.py
tcouch: Yes, but I had to fix a few things to do with testimonium ordering: basically put testimonia belonging to unknown works before those attached to other works (opposite to how it works for fragments and apposita)

Issue #380, opened 2023-03-14
Moving links between works/books


acholyn:
Currently the method to update a link's work/book is to create a new one with the right information. A solution could be to allow the user (in edit mode on the work/book page) to change this via an inline form with the dropdowns for work/book and whether this new allocation is definite or possible (see #379)

rmamarshall:
@acholyn @tcouch Hi, to clarify my comments earlier, there are potentially several (possible) components that constitute a fragment's identity:

1. Possible Antiquarian
+ Possible work 
++ Possible book
++ Definite book [we may know that the fragment belongs to book three, but not the title of the work!]
OR
+ Definite work [we may know the title of the work, but not the author]
++ Possible book [we may know the fragment belongs to a work, but not which book of that work]
++ Definite book

2. Definite Antiquarian
+ Possible work 
++ Possible book
++ Definite book [we may know that the fragment belongs to book three, but not the title of the work]
OR
+ Definite work
++ Possible book [we may know the fragment belongs to a work, but not which book of that work]
++ Definite book

acholyn:
would that mean if the link has a definite work/book you wouldn't want it to be possible to change it?

rmamarshall:
@acholyn. If the Antiquarian is also definite, yes, it should not be possible to change the location of a fragment with definite work / book. 
If the antiquarian is only possible, but the work is definite, this can only mean that there are two works by two different authors that share the same title in the database, and we don't know to which one the fragment should be assigned. This also should not be changed _within_ an antiquarian entry.
If only a book is definite, it should only be possible to place the fragment in e.g. book three of any given work (though we might want to place it in two or more works simultaneously). I hope this makes sense!

The possible scenarios are:
1. Varro in book 2 of the _Disciplines_ says ...
Definite Antiquarian, Definite work, definite book.

2. Varro in the _Disciplines_ says ...
Definite Antiquarian, Definite work, unknown book (though we may was to assign a one or more possible books to the fragment).

3. Someone in book 2 of the _Disciplines_ says ...
This could be Varro or Nepos (so the fragment has two possible antiquarian identifications), but the work and book is definite, so can only appear in Book 2 of the Disciplines of Nepos and / or Book 2 of the Disciplines of Varro.

4. Someone in the _Liberal Arts_ says ...
[...]

PR #389, opened 2023-05-30
Feature/definite possible clarity

### LINKED ISSUE #379: Definite/Possible Clarity
Currently links have a definite/possible attribute which might refer to a definite link to an Antiquarian and a work/book or just the Antiquarian. To give better visibility on which links are known to belong to an antiquarian/work/book. This could be done by adding new attributes for each link, being `definite_work` and `definite_book` and rename the current to `definite_antiquarian`. This could then be represented by different symbols/colours for the statuses.

acholyn:
Closes #379 
Also related to #380 

Things changed:

- Migration to add in new `definite_antiquarian`, `definite_work` and `definite_book` fields for links, with the former `definite` field values being assigned to the new `definite_antiquarian` field
- All links now have symbols to represent whether they are definitely linked to the antiquarian/work/book
- On Fragment/Testimonium: have organised links so they are grouped by antiquarian with all relevant links below
    - When removing links, one can remove a single link with the button next to the work/book link; if it is the last link for that antiquarian it will reassign itself to unknown work & book. 
    - If the last link is unknown, the link must be removed from the button by the antiquarian, which removes all links between that antiquarian and the fragment/testimonium
- It should not be possible to indicate that a link has a definite work/book if that work/book is unknown
- New inline update form for links - book can be changed as well as definite statuses
- Tests have been updated to reflect the new deletion behaviour expected as well as the different definite aspects and ensuring unknown works/books can't have definite links

Other updates: more relevant tab titles


src/rard/templates/research/partials/linked_antiquarian.html
```diff
@@ -22,17 +28,17 @@
...
             {% else %}
-            <form novalidate class='form-inline' action='{% url "testimonium:remove_testimonium_link" link.pk %}' method='POST'>
+            <form novalidate class='form-inline' action='{% url "testimonium:remove_testimonium_link" antiquarian.pk %}' method='POST'>
             {% endif %}
                 {% csrf_token %}
+                 <input type="hidden" name="antiquarian_request" value={{object.pk}} />
                 <div class='form-group'>
                     <button type='submit' class='btn btn-link text-danger confirm-delete p-0 ml-2'
-                        data-what='link'>{% trans 'Remove Link' %}</button>
+                        data-what='antiquarian_link'><small>{% trans 'remove all links' %}</small></button>
```
tcouch: I think this button needs to come with a warning: "Are you sure you want to delete all links associated with this antiquarian and their works". In the same way you get a warning for Make Anonymous, or Make Unlinked

src/rard/templates/research/add_work_link.html
```diff
@@ -9,25 +14,37 @@
...
     </form>
 
+    {# work form #}
     <form novalidate enctype="multipart/form-data" autocomplete='off' action='{{ request.path }}' class="form work-form" method='GET'>
         {% bootstrap_field form.work field_class='submit-on-change' %}
+        <span class="toggle-definite">
+        {% bootstrap_field form.definite_work %}</span>
         <input type='hidden' name='antiquarian' value='{{ antiquarian.pk }}' />
-    </form>
+        <input type='hidden' name='definite_antiquarian' {% if definite_antiquarian %} value={{ definite_antiquarian }} {% endif %} />
```
tcouch, line 31: If I select definite antiquarian after first page reload, then select a work, the definite antiquarian box is unchecked when the page reloads again. Would it be possible to add some JS to update this input value when user changes definite checkboxes?

src/rard/templates/research/add_work_link.html
```diff
@@ -9,25 +14,37 @@
...
 
+    </form>
```
tcouch, line 33: Also, definite values aren't getting carried through when I click submit.

src/rard/templates/research/partials/render_inline_worklink_form.html
```diff
@@ -0,0 +1,23 @@
...
+<div class="inline-worklink__form" hx-target="this" hx-swap="outerHTML">
+    <form novalidate enctype="form-data" autocomplete='off' action='{{ request.path }}' hx-post='{% url inline_update_url link.pk %}'>
+    {% csrf_token %}
+    <div class="inline-worklink__form-field">
+       <p> {{form.instance.antiquarian}}</p>
+        {% bootstrap_field form.definite_antiquarian field_class="toggle-definite" %}
+    </div>
+    <div class="inline-worklink__form-field">
+        <p>{{form.instance.work.name}}</p>
+        {% bootstrap_field form.definite_work field_class="toggle-definite" %}
```
tcouch, line 14: This form is letting me update antiquarian definite status and change book, but definite work and definite book don't stick.

tcouch (changes requested):
I've not gone through everything, but found a few things aren't working as expected when creating/updating links.

src/rard/templates/research/partials/linked_antiquarian.html
acholyn: oki - does it come up with any alert at all? It has the confirm-delete class so it _should_ but I also noticed it didn't work for me and then it started working for me without me changing the code... 
That might be something to look into if you're not seeing it pop up

src/rard/templates/research/add_work_link.html
acholyn: We do have that, but when you select an antiquarian it then submits the form, so you're doing it in the limbo of the page refresh. I'll check what's going on as it used to function
Should be lines 57-72 for the script

src/rard/templates/research/add_work_link.html
acholyn: This is unexpected; I'll get back to you

src/rard/templates/research/partials/render_inline_worklink_form.html
acholyn: Also unexpected since it was working before; I'll see what's happening on my end

tcouch (approved):

Issue #376, opened 2023-01-30
Adding intro box to works/books


acholyn:
Related to #375 ; mentioned in #370

- [x] add intro text box to each work and book - appears before children
- [x] make intro box editable from work page

tcouch:
Can we have option to create introduction for books, but not have them by default? Always have intro for works.

acholyn:
both are optional

PR #394, opened 2023-06-21
Feature/add intro boxes

acholyn:
Summary:
Have added introduction boxes to works and books. This involved changes to the templates, forms, models and views (and migration). Related to #376 

Models
- have added introduction - 1:1 relationship with TextObjectField instance
- have added plain introduction - plain text version for search purposes
- this should be the same across ant/work/book
- when a new work/book is created, it should create an introduction (TOF) object associated with it

Views
- new update introduction views based on update views- the work one follows the same pattern as the ant one in which it just uses the detail page with functions for handling an update. The book one is more complicated as books have no detail page so it redirects to the creation form but will act as an update. There is a filler placeholder (this is empty) that will create a text object associated with the work/book if it does not already exist

Forms
- Have taken the existing antiquarian introduction form and created a base form which the ant/work/book introduction forms are based on
- have added introduction fields to book and work (create) forms

Templates
- have added the boxes - which just makes use of the preexisting `text_object_preview.html` partial, same situation for the rich text editors for adding/updating the introduction content
- have added add/edit buttons depending on if the introduction exists (has not been created yet) and if it is less than a certain length (the content at base has some empty tags - this might be a good idea to update)
- similarly to above, the boxes showing the introductions should not show if they are empty
- with the new editing states, they also have tab descriptions

Tests
- check that `TextObjectField` instances associated with the work/book introduction are created/deleted with the work/book (models)
- check that the content is properly updated and that it is in the editing mode (views)


src/rard/research/models/work.py
```diff
@@ -49,6 +52,17 @@ class Meta:
...
+        "TextObjectField",
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="introduction_for_%(class)s",
+    )
+    plain_introduction = models.TextField(blank=False, default="")
+
+    def book_set(self):
+        return super().book_set().order_by("unknown", "order", "number")
+
```
tcouch: I don't understand what this is for. The meta class on the Book model already orders books by unknown, order, number... so I don't think this actually does anything.

src/rard/research/models/text_object_field.py
```diff
@@ -22,8 +22,10 @@ def related_lock_object(self):
     references = GenericRelation("BibliographyItem", related_query_name="text_fields")
 
     def get_history_title(self):
+        from rard.research.models import Book, Work
+
```
tcouch: Can get class name as string here to save importing models:
```suggestion
        obj = self.get_related_object()
        if obj.__class__.__name__ in ["Antiquarian", "Work", "Book"]:
```

tcouch: Annoying, suggestion doesn't work properly because it's across a deleted line, but you get the gist.

src/rard/research/migrations/0063_work_book_introductions.py
```diff
@@ -0,0 +1,76 @@
...
+            name="introduction",
+            field=models.ForeignKey(
+                null=True,
+                db_constraint=False,  # Adding this manually to prevent error when deleting works
+                on_delete=django.db.models.deletion.SET_NULL,
+                related_name="introduction_for_historical_work",
+                to="research.textobjectfield",
+            ),
+        ),
+    ]
```
tcouch, line 76: So, when I started this up on my machine, django logs said there were changes to models that needed migrating. Make migrations brought up changes to Antiquarian.introduction, HistoricalBook.introduction, and HistoricalWork.introduction. I get the Antiquarian one because related name has changed, but historical changes are a mystery. In any case, adding 0064 migration should suffice.

src/rard/research/models/work.py
acholyn: Ah, this shouldn't be here. This branch was created off the definite/possible before it was merged and before this had been deleted. Will delete, thank you for spotting it

src/rard/research/migrations/0063_work_book_introductions.py
tcouch: We should probably have that as a CI test: i.e. it fails if running makemigrations creates a new migration file. I'll create an issue.

src/rard/research/views/work.py
```diff
@@ -83,18 +88,48 @@ def get_success_url(self, *args, **kwargs):
...
+            to_add = [a for a in updated if a not in existing]
+            for a in to_remove:
+                a.works.remove(work)
+            for a in to_add:
+                a.works.add(work)
+            add_new_books_to_work(work, form)
         return super().form_valid(form)
 
 
+class WorkUpdateIntroductionView(WorkUpdateView):
```
tcouch, line 104: I'm seeing this when I click Add Introduction on the Work edit page:
image

tcouch: Same for the book

tcouch: Ok, ignore that, it was me. Privacy Badger had identified Quill.js as a potential tracker

src/rard/research/forms.py
```diff
@@ -293,11 +309,44 @@ def clean(self):
...
+        if commit:
+            instance.save_without_historical_record()
+            # introduction will have been created at this point
+            instance.introduction.content = self.cleaned_data["introduction_text"]
+            instance.introduction.save_without_historical_record()
+        return instance
+
 
 class BookForm(forms.ModelForm):
+    introduction_text = forms.CharField(
```
tcouch, line 323: Some code repetition here: could you have created a mixin?

src/rard/research/models/text_object_field.py
```diff
@@ -42,6 +41,12 @@ def get_related_object(self):
                 pass
         return None
 
+    def save(self, *args, **kwargs):
```
tcouch, line 44: Can you add a comment here to explain why we save the parent object if it's Antiquarian, Work or Book (and why not if it's Fragment or whatever)

src/rard/research/views/work.py
```diff
@@ -83,18 +88,48 @@ def get_success_url(self, *args, **kwargs):
...
+    template_name = "research/work_detail.html"
+
+    def create_intro_if_does_not_exist(self, *args, **kwargs):
+        work = self.get_object()
+        if work.introduction is None:
+            work.introduction = TextObjectField.objects.create(content="")
+            work.save()
+
+    def dispatch(self, request, *args, **kwargs):
+        self.create_intro_if_does_not_exist()
```
tcouch, line 120: Could you add a comment to explain why we need to do this?

src/rard/research/views/work.py
```diff
@@ -83,18 +88,48 @@ def get_success_url(self, *args, **kwargs):
...
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_success_url(self, *args, **kwargs):
```
tcouch: This is almost identical to the get_success_url function in the parent WorkUpdateView class. Is this necessary?

src/rard/research/views/work.py
```diff
@@ -183,6 +217,39 @@ def get_context_data(self, *args, **kwargs):
...
+    template_name = "research/book_form.html"
+
+    def create_intro_if_does_not_exist(self, *args, **kwargs):
+        book = self.get_object()
+        if book.introduction is None:
+            book.introduction = TextObjectField.objects.create(content="")
+            book.save()
+
+    def dispatch(self, request, *args, **kwargs):
+        self.create_intro_if_does_not_exist()
```
tcouch, line 236: As above, could you add a comment here to explain why we need to do this?

src/rard/research/views/work.py
```diff
@@ -183,6 +217,39 @@ def get_context_data(self, *args, **kwargs):
...
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_success_url(self, *args, **kwargs):
```
tcouch: And, as above, do you need to override this method?

src/rard/research/views/work.py
```diff
@@ -183,6 +217,39 @@ def get_context_data(self, *args, **kwargs):
...
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context.update(
+            {
+                "editing": "introduction",
+                "change_work": True,
+                "change_book": True,
+                "has_object_lock": True,
+                "instance": self.get_object(),
```
tcouch: What are these for? The only one I can see being used in the template is has_object_lock and usually that's worked out in the template as
`{% with request.user|has_lock:object as has_object_lock %}`

tcouch (approved):

src/rard/research/forms.py
tcouch: Not a request, just throwing it out there

acholyn: I hadn't thought of that - I would like to go through and streamline things at some point actually (there's a fair bit of repetition spread around). Would it be worth opening an issue for? There's one for beautifying the db but that's not quite the same

src/rard/research/models/text_object_field.py
acholyn: have tried - let me know if I'm wrong or if it just needs some rewording

tcouch: Your comment makes me think though, shouldn't we save the related object whatever type it is? Assuming this is to do with making sure plain_introduction gets updated for the purpose of searching, we'd also want plain_commentary to be updated for Fragments.

src/rard/research/views/work.py
acholyn: since the book update redirects to a new page (there's no book detail page) it loses some of the context (like editing permissions)
it's possible that this is no longer the case - I can try removing parts and seeing if it can be slimmed down

src/rard/research/views/work.py
acholyn: as it turns out, we don't 👍

src/rard/research/models/text_object_field.py
acholyn: hmmm...surely there's an explanation..?
I guess I don't really know _why_ we don't save all of them or why the commentary works without the extra save 🤔

tcouch: Maybe it's not possible to edit commentary without also saving the fragment via the UI? If you edit the commentary content via the shell and save it, does the fragment's plain_commentary get updated?

PR #397, opened 2023-07-14
Fix multiple unknown books

tcouch:
Fixes the issue where a new unknown book is added every time a work is saved.


acholyn (approved):
Also did this on the text editor branch but hopefully that doesn't cause any merge issues

Issue #392, opened 2023-06-13
Adding footnotes to introduction boxes


acholyn:
Previously, the researchers were told this wasn't possible due to the nature of the text editor. This should be investigated to find out if there is a way to implement this as it is a text-only feature - no additional functionality needed.
@ mentions would not need to work in the footnotes

Might overlap with #319 
As #71 also requests a text feature there might be reason to choose a new editor

acholyn:
It seems [QuillJS](https://quilljs.com/) is responsible for the rich text editor. Whilst it doesn't automatically come with the footnote option it may be possible to build a custom module to handle this

Issue #319, opened 2022-04-04
Text editor add space between lines in view mode


raquelalegre:
Reported by @rmamarshall. When creating an original text in edit mode, a user sees a spacing between lines that looks reasonable. However, upon saving and being presented with the text in view mode, the editor seems to have added increased the space between each line. The team prefers to maintain the spacing between lines the edit mode shows.

acholyn:
If this is still the case, would it be possible for someone to share a screenshot as the spacing looks the same on my own laptop

rmamarshall:
@acholyn 
![image](https://github.com/UCL/frrant/assets/75932472/38f84e62-7296-46f5-8fd6-79f7918f98ce)
![image](https://github.com/UCL/frrant/assets/75932472/3069b34a-1f0b-4036-8f1f-980b96c1a411)

Between the first two lines and the rest of the passage, a space is added once the text editor (first image) is closed. The second line ends in a hard return. This automatic formatting should be removed, as many kinds of evidence, including poetry (every verse of which ends in a hard return) is currently displayed incorrectly.

rmamarshall:
@acholyn Hi - realised I've mixed screenshots from the Original Text and Translation boxes, but the issue is the same in all text editor boxes - once text is saved, the automatic formatting function adds inappropriate spaces after hard returns.

acholyn:
How's this?
<img width="422" alt="image" src="https://github.com/UCL/frrant/assets/72266023/8b2b5567-c3d5-4bec-aa9c-a2409fcdd821">

Issue #71, opened 2020-12-01
Add columns to texts when editing


raquelalegre:
Rich text editor functionality should include writing in columns and tables.

@peterbispham has tested tables and they don't work well in the current editor we are using. Any other possible solutions?

tcouch:
To follow up with @rmamarshall (one solution might be to add multiple boxes side by side)

raquelalegre:
@rmamarshall do you know how many columns typically? Can we get away assuming there's 2 columns?

rmamarshall:
@raquel-ucl @tcouch As promised in the meeting on Friday, here is an example of the type of material for which we need this feature. In the attached example, there are five columns. Note, however, that in some rows, the final pair of columns is merged: a table feature might be the ideal solution, though I have no idea how feasible this is. 
I've been thinking about the suggested work-around (multiple text boxes): this could also work, but I presume the app. crit. would be discrete for each box, which would be a problem if we ended up with multiple notes relating to the same original text that shared the same app. crit. number.

![Jerome_Chron_Columns](https://user-images.githubusercontent.com/75932472/140764408-4ae993bc-8c5e-4b47-9c53-bda770c8779f.JPG)

tcouch:
I've added some information on how the rich text editor works here: https://github.com/UCL/frrant/wiki/Notes-on-FRRAnt-Rich-Text-Editor

tcouch:
There appear to be a few options to try:
* Quilljs version 2.0.0 includes a table module which may do the job. Development seems to have stopped though before an official release so it's not a sure bet and there may be breaking changes to existing functionality that we rely on.
* We can look for an alternative rich text editor. If we find one that provides everything we need, we'll have to convert all our mentions etc. to whatever format the new editor expects them to be in.
* A slightly hacky work around could be for users to enter tables in markdown format and convert these to html when we validate the form data. This breaks the WYSIWYG principle though and puts the burden of learning markdown tables on the user.
* There's also the multiple editors option proposed above, but I'm not a fan.

acholyn:
@rmamarshall Revisiting this - is it tables _and_ columns you'd like or is it mainly tables due to how you'd input the information? It _might_ be possible to include both if you want both for different uses, but the above example seems more along the lines of a table.

rmamarshall:
[...]

Issue #53, opened 2020-12-01
Vincula seems broken with new font


raquelalegre:
Needs to be shown on top of the letters, but at the moment it shows in the middle of them.

tcouch:
This is going to be important when working with Roman numerals in particular. At the moment, vincula over a word appear as a non-continuous line.

acholyn:
fixed with styling - see upcoming PR

PR #401, opened 2023-08-03
Feature/text editor updates

acholyn:
This addresses issues #392, #53, #319 & #71.

Small fixes for #53 and #319: styled the unicode icon for the vincula so the lines do not appear broken when over consecutive characters; reduced the margin and padding of the `text-object-preview` elements so the spacing is similar to the editor.

Footnotes (#392 )
- these can be added with the comment button in the editor
- it will add an indicator in the main text section and a note in the notes section at the bottom of the editor
- you should be able to see a note's contents by hovering the indicator (if these are likely to be very long this can be updated so it's truncated)
- if you want to edit the contents of the note you can click the note in the notes section at the bottom
- you can delete the note (and it's corresponding indicator) by removing the contents of the note or backspacing the indicator
- the notes and indicators should update their numbers when adding or removing new ones
- it should be possible to delete more than one note at once (by highlighting and deleting a section of text containing more than one indicator)
- there are footnote hints in a tooltip at the bottom of the editor

Tables (#71)
- I have implemented tables based on the discussion in the issue, as opposed to columns
- QuillJS 2.0 (not yet released) has a table module which I have adapted and stripped back for our use case
- functions available: adding a 3x2 table as base (this can be updated if there's a another size that would be preferred), adding rows (always to the bottom), adding columns (always to the right), removing columns and rows selectively
- due to how the table is added to the editor, be careful when editing contents! I've made it so that you double click on a cell to edit its contents because if you backspace, it will delete the table (no bueno)
- in the editor, the grid lines are visible but not on the preview (this can be changed if preferred)
- the cells will resize based on content
- you should always be able to add a new column or row, even if there's content or your cursor isn't on the table - in the case of the latter it will perform the edit on the closest table
- there are hints at the bottom of the editor, in a tooltip


src/rard/static/js/project.js
```diff
@@ -177,16 +177,18 @@ $("body").on("submit", "form", function (e) {
   $(this).data("submitted", true);
 });
 
-Quill.register("modules/mention", quillMention);
+Quill.register("modules/mention", quillMention, true);
 var icons = Quill.import("ui/icons");
-// import fontawesome icons for the undo/redo buttons
+// import fontawesome icons buttons
```
tim-band: 'import fontawesome for button icons' makes a bit more sense.

PR #405, opened 2023-08-22
Feature/extra editor updates

acholyn:
This is an expansion of the features created in #401 and addresses aspects of #392 and #71 

Features include:
- replacing most prompts with modals containing text editors
  - tables nested within footnotes and other tables will not trigger a modal as this breaks the editors
- footnotes have been refactored so they function within their editor --- there was an issue with the numbers due to the editing being made inline
- have removed the ability to try nest footnotes (breaks editors)
- ensuring symbol picker works with the modal open --- moving it within the DOM
- ensuring mentions and apparatus critici work in the modals

Somewhat unrelated but was triggered by a production error: 
- have removed the ability to edit unknown books (breaks)


tcouch:
One thing I've noticed is that backspacing the footnote indicator doesn't delete it from the footnotes section. Did that used to work?

tcouch:
Can we set a minimum width for table columns? Here both cells' contents are on one line, but the smaller one is forced to wrap:
image

tcouch:
The hint text extends over the button here, making just the bottom of it clickable
image

src/rard/research/views/home.py
```diff
@@ -28,3 +29,22 @@ def get_queryset(self):
             .annotate(topic=F("topics__name"))
             .order_by("topic", "topiclink__order")
         )
+
+
+def render_editor_modal_template(request):
+    import re
+
+    from rard.utils.context_processors import symbols_context
+
```
tcouch: Can you add a comment here to describe what's happening, with an example?

src/rard/static/js/fetchAndAddEditorModal.js
```diff
@@ -0,0 +1,74 @@
...
+  var modalHTML = await getModalTemplate();
+  modalContainer.innerHTML = modalHTML;
+  return new Promise((resolve) => {
+    document.body.appendChild(modalContainer);
+    $(modalContainer).find(".modal").modal({
+      backdrop: "static",
+    }); // init the modal
+    // init the editor
+    initRichTextEditor($("#modal-content-editor"));
+    setTimeout(() => {
```
tcouch, line 38: what does the setTimeout bit do?

tcouch (approved):

src/rard/static/js/fetchAndAddEditorModal.js
acholyn: it makes sure that the modal doesn't try initialise immediately since the ql-editor needs to initialise first (as it needs to select for it). the way the editors are initialised means I couldn't assign it to a variable and say `if editor variable...`

acholyn:
> One thing I've noticed is that backspacing the footnote indicator doesn't delete it from the footnotes section. Did that used to work?

It did...I wonder what's happened to stop that...

> Can we set a minimum width for table columns? Here both cells' contents are on one line, but the smaller one is forced to wrap: image

Can do, I think actually a max might help as well

src/rard/research/views/home.py
acholyn: have actually gutted this since the way I ended up doing it doesn't require the context to be passed anymore in my testing. Still need the view to generate the url to fetch afaict

Issue #378, opened 2023-03-14
Order anonymous section by citing author then reference number within topic


tcouch:
When users choose to display anonymous fragments ordered by reference number, this should be by citing author and then reference number.

E.g.:
Livy 1.1; Gell. 1.2; Livy 1.3.
Should actually be displayed as:
Livy 1.1; Livy 1.3; Gell. 1.2.

See Anonymous Fragments: Etymology for example.

(created from #290)

Issue #286, opened 2021-11-04
Title for site should start with "The"


tcouch:
Title and heading in banner should be "The Fragments of the Roman Republican Antiquarians Research Database"

Issue #168, opened 2021-02-16
Styling improvements


raquelalegre:
Change colouring scheme and style using backgrounds like the ones in https://www.toptal.com/designers/subtlepatterns

acholyn:
going to close this as this was done, in favour of using #371 for other, more specific, styling tweaks

PR #406, opened 2023-08-31
Fix/small updates

acholyn:
Addresses a few small issues:
#286 #378 #168
- added in some more page titles to help with navigating tabs
- created a function that runs on ant/work save to collate contents of unknown works/books if there's more than one (previously there was a bug that kept creating unknown works/books)

new styling:
image

### DIFF
--- a/src/rard/utils/shared_functions.py
+++ b/src/rard/utils/shared_functions.py
@@ -42,41 +42,17 @@ def reassign_to_unknown(worklink):
     worklink.save()
 
 
-def collate_unknown(instance):
-    """This makes sure there's only one unknown work/book per antiquarian/work and combines contents if otherwise"""
-    if instance.__class__.__name__ == "Antiquarian":
-        unknown_works = instance.works.filter(unknown=True)
-        if unknown_works.count() > 1:
-            designated_unknown = unknown_works.first()
-            other_unknown_works = unknown_works.exclude(pk=designated_unknown.pk)
-
-            for uw in other_unknown_works:
-                transfer_links(instance.fragmentlinks.all(), uw, designated_unknown)
-                transfer_links(instance.testimoniumlinks.all(), uw, designated_unknown)
-                transfer_links(
-                    instance.appositumfragmentlinks.all(), uw, designated_unknown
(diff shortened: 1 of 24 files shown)


src/rard/utils/shared_functions.py
```diff
@@ -40,3 +40,46 @@ def reassign_to_unknown(worklink):
...
+                )
+                transfer_links(ub.testimoniumlinks.all(), ub, designated_unknown)
+                transfer_links(
+                    ub.antiquarian_book_appositumfragmentlinks.all(),
+                    ub,
+                    designated_unknown,
+                )
+
+            other_unknown_books.delete()
+
```
tim-band: Do you want to catch other cases (assuming you still want this iffy way to do it)? Seems like any other case would be a programming error. Should probably throw an exception in this case, or otherwise `logging.error("cannot collate unknown books of class %s", instance.__class__.__name__)` or something

src/rard/utils/shared_functions.py
```diff
@@ -40,3 +40,46 @@ def reassign_to_unknown(worklink):
     worklink.definite_work = False
     worklink.definite_book = False
     worklink.save()
+
+
+def collate_unknown(instance):
+    """This makes sure there's only one unknown work/book per antiquarian/work and combines contents if otherwise"""
+    if instance.__class__.__name__ == "Antiquarian":
```
tim-band: iffing on class name? Strongly consider adding a `collate_unknown` method to Antiquarian and Work model objects instead.

src/rard/utils/shared_functions.py
```diff
@@ -40,3 +40,46 @@ def reassign_to_unknown(worklink):
...
+        unknown_works = instance.works.filter(unknown=True)
+        if unknown_works.count() > 1:
+            designated_unknown = unknown_works.first()
+            other_unknown_works = unknown_works.exclude(pk=designated_unknown.pk)
+
+            for uw in other_unknown_works:
```
tim-band: I would probably extract this into a separate function, but it's up to you.

PR #415, opened 2024-01-16
Feature/move links between works

### LINKED ISSUE #380: Moving links between works/books

acholyn:
This update allows researchers to reassign fragment/testimonium links to works and books by the same antiquarian without having to create a new link and delete the old one.
closes #380


src/rard/templates/research/partials/render_inline_worklink_form.html
tcouch, line 1: I think this should go in `templates/research/inline_forms`

src/rard/research/forms.py
```diff
@@ -805,11 +806,15 @@ def __init__(self, *args, **kwargs):
         if update:
             antiquarian = kwargs["initial"]["antiquarian"]
             work = kwargs["initial"]["work"]
+            if "data" in kwargs:
+                # this is where the new values are
+                work = Work.objects.get(pk=kwargs["data"]["work"])
+                book = Book.objects.get(pk=kwargs["data"]["book"])
```
tcouch: Can you check if a book id is returned first?

tcouch (changes requested):
At the moment, if a user clicks update on the form, but hasn't selected a book, nothing appears to happen. Behind the scenes you get an error message because not book can be found. I think either the form should be returned with a helpful error message, or the submit button should be disabled until both a work and book have been selected.

acholyn:
ah - how it worked previously was that if no book was selected it automatically assigned the unknown one of that work. I didn't realise this had stopped being the case - should I make it so again? If not I'll disable the submit

tcouch:
Yeah, defaulting to unknown sounds sensible. I suppose in that case, there's no reason to have the none option?

acholyn:
I think without having the blank it chooses the first option (which would be an actual book) so that wouldn't work. I could change it to say unknown? It wouldn't have a value and would mean unknown shows up twice (if I remember correctly)

tcouch (approved):

PR #416, opened 2024-01-22
Fix/414

### LINKED ISSUE #414: Appositum fragment link form needs a filter
The form to add a fragment link to an anonymous fragment is a simple drop down presenting all fragments which is very difficult to use now that the database has grown.

A solution would be to provide filters - Antiquarian and Work - in order to narrow down the number of items in the drop down. It's important that the results are also in numerical order as that will help quickly identify the right item.

acholyn:
Closes #414 
----
Have changed the appositum-fragment linking form to search and then populate a dropdown, whose value gets submitted as previously. The search is triggered on keyup with a delay of 500ms. It utilises the preexisting functions within the `MentionSearchView` to search for the normal or unlinked fragments.

### DIFF
--- a/src/rard/research/views/fragment.py
+++ b/src/rard/research/views/fragment.py
@@ -41,6 +41,7 @@ from rard.research.models import (
 )
 from rard.research.models.base import AppositumFragmentLink, FragmentLink
 from rard.research.models.fragment import AnonymousTopicLink
+from rard.research.views.mention import MentionSearchView
 from rard.research.views.mixins import (
     CanLockMixin,
     CheckLockMixin,
@@ -988,3 +989,17 @@ class MoveAnonymousTopicLinkView(LoginRequiredMixin, View):
                 raise Http404
 
         raise Http404
+
+
+def fetch_fragments(request):
+    query = request.GET.get("search-fragments")
+    if "unlinked" in query:
+        fragments = MentionSearchView.unlinked_fragment_search(query)
+    else:
+        fragments = MentionSearchView.fragment_search(query)
+
+    return render(
+        request,
+        "research/partials/htmx_fragment_results.html",
+        {"fragments": fragments},
+    )
--- a/src/rard/research/views/fragment.py
+++ b/src/rard/research/views/fragment.py
@@ -992,11 +992,17 @@ class MoveAnonymousTopicLinkView(LoginRequiredMixin, View):
 
 
 def fetch_fragments(request):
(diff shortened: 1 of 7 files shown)


src/rard/templates/research/add_appositum_fragment_link.html
```diff
@@ -5,20 +5,23 @@
...
         Add appositum link to fragment
     {% endblocktrans %}
+    {{ form.media.css }}
 {% endblock %}
 
 {% block inner %}
 
     <form novalidate enctype="multipart/form-data" autocomplete='off' action='{{ request.path }}' class="form" method='POST'>
         {% csrf_token %}
+        <input type="text" id="search-fragments" name="search-fragments" placeholder="Start typing to search for fragments..." hx-get="{% url 'fragment:fetch_fragments' %}" hx-trigger="keyup delay:500ms changed" hx-swap="innerHTML" hx-target="#search-results" class="form-control">
```
tcouch, line 15: This element seems a bit close to the one below, could maybe do with a bigger margin?

src/rard/research/views/fragment.py
```diff
@@ -988,3 +989,23 @@ def post(self, *args, **kwargs):
                 raise Http404
 
         raise Http404
+
+
+def fetch_fragments(request):
+    # Mention search methods expect a list of search terms so split on space
+    query = request.GET.get("search-fragments")
```
tcouch, line 996: I updated this to do two things:
1. make an empty query quickly return none
2. the mention search methods expect a list of search terms, if they're passed a single string, they'll search for each character separately (one of the joys of loose typing)

src/rard/templates/research/partials/htmx_fragment_results.html
```diff
@@ -0,0 +1,7 @@
+{% if fragments %}
+  {% for frag in fragments %}
+        <option value="{{frag.pk}}">{{frag}}</option>
+  {% endfor %}
+{% else %}
+  <option value="" selected="">Results will appear here</option>
```
tcouch, line 6: Resets the select options if there's an empty search result

tcouch (changes requested):
If you look at the search methods they check to see if any of the search terms are an integer, in which case they use this to add an order_number query. Users can take advantage of this by typing something like "publius 12" to narrow down the search. Can you add some help text to let users know?

tcouch (approved):

PR #417, opened 2024-01-23
adding underdot same way as vincula

### LINKED ISSUE #369: Some characters with ‘uncertainty dot’ missing from symbol picker
We use the characters with a dot below (e.g. ‘ạ’, ‘Ạ’) to indicate that a letter is not clearly legible in a manuscript or inscription. However, some letters with dots are missing from the symbol picker. I don’t know if they’re in the Alphabetum font and can easily be added to the symbol picker, or if they’re missing from the font altogether, in which case we would probably have to draw them ourselves using the system Peter devised for adding new characters to the font.

In Latin, the following letters with dot below are missing from the symbol picker: c, C, f, F, g, G, p, P, q, Q, and x. (Uppercase X with a dot is already there, presumably because it’s identical to Greek uppercase χ with a dot, = EA8E.)

In Greek, we’re missing only lowercase ξ with dot below (capital Ξ with dot is there, = EA56).

acholyn:
closes #369
---
Underdot character has been added to the text editor in the same way as vincula.

### DIFF
--- a/src/rard/static/js/project.js
+++ b/src/rard/static/js/project.js
@@ -187,6 +187,8 @@ icons["redo"] =
 // and for the custom buttons
 icons["vinculum_on"] = "V\u0305";
 icons["vinculum_off"] = "V";
+icons["underdot_on"] = "U\u0323";
+icons["underdot_off"] = "U";
 icons["footnote"] = '<i class="far fa-comment-alt"></i>';
 icons["table"] = '<i class="fas fa-table">+</i>';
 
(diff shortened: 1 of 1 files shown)


tcouch (approved):
One thing I've found is that underdot_off doesn't always work. It seems to depend on which character it has been applied to. Adding underdot to any greek characters seems fine (you can also remove it with backspace), but this doesn't work for most english alphabet characters. In fact, adding an underdot, then pressing backspace deletes the following character.

If I had to guess I'd say that quill is automatically converting combined character combos to their single character equivalent where possible, but the length gets messed up somehow so backspace ends up referring to the following character.

Since this seems like a quill problem I'm not sure it's worth spending too much time trying to fix it. The research team specifically mentioned they'd like to be able to add underdots to greek characters and that's working fine. So while this is a bug which is relatively easy to encounter, I don't think it's stopping anyone from doing what they need to do, and I don't think it's going to confuse users too much.

What do you think?

acholyn:
I didn't notice that behaviour during my usage - I wonder if that's something to do with browsers/OSes as I just checked and had no issues with regular alphabet characters nor with greek characters. I think it should be ok for the researchers and if they find some particularly buggy behaviour then we can look into it.

Issue #358, opened 2022-09-22
When deleting a fragment, alert users if there are @mentions referencing it


tcouch:
When deleting a fragment, alert users if there are @mentions referencing it

acholyn:
Is this meant to prevent the deletion or just provide a notification? Should this also include testimonia since that tracks mentions as well

acholyn:
disable deletion: can only delete if no mentions

PR #418, opened 2024-01-25
Feature/track ft mentions

### LINKED ISSUE #390: Can we track where items are mentioned and display list at bottom of the page?
Related to https://github.com/UCL/frrant/issues/358

It would be most helpful to see where (Anonymous/Unlinked) Fragments and Testimonia are mentioned in other introductions and commentaries

This may assist with #296

acholyn:
closes #390 
---
When a `TextObjectField` (`TOF`) instance is saved, a function will run to `update_mentions`, which involves calling another function that turns the content into beautiful soup and searching for all `<class="mention">` elements and then filters these so only the ones for `fragment`, `anonymousfragment` and `testimonium` are processed. 
The list of mentioned objects is returned to `update_mentions` and this list is compared to the preexisting list of mentioned objects called `fragment_testimonia_mentions`. 
If the lists do not differ, nothing is updated. If they do differ, `set_fragment_testimonia_mentions` is called with the list of items to add and a list of items to remove from `fragment_testimonia_mentions`. The items in the removal list are removed from the mentioned object's `mentioned_in` set and are removed from the `TOF.fragment_testimonia_mentions`, the reverse is true for items in the "to add" list.

Upon completion, each `TOF` should have a list of `fragment_testimonia_mentions` and each of these items should have the parent/owner object of the `TOF` in their `mentioned_in_list` (a property which extracts the related object of the `TOF`s in the `mentioned_in` set.

This should act as a sensible foundation for #358


tcouch:
The individual items in the mentions list could do with something to make it clearer they're separate items rather than one item with a long name - maybe bullet points, maybe just spacing?
image

src/rard/research/forms.py
```diff
@@ -655,6 +655,9 @@ class Meta:
 
 
 class HistoricalFormBase(forms.ModelForm):
+    class Meta:
+        readonly_fields = ("mentioned_in",)
```
tcouch: I don't think this does anything. There's no reference to this option in the documentation and I can't find anything in the source code for ModelForm: https://github.com/django/django/blob/main/django/forms/models.py

src/rard/research/migrations/0067_add_mentioned_in_fk_field.py
```diff
@@ -0,0 +1,88 @@
...
+            try:
+                model = apps.get_model(app_label="research", model_name=model)
+                m = model.objects.get(pk=int(pk))
+                linked_items.append(m)
+            except ObjectDoesNotExist:
+                pass
+    return linked_items
+
+
+def set_mentions(self, apps):
```
tcouch: I wouldn't use self as the argument name when it's not a method belonging to a class. Why not text_object like when you call it in update_mentions? Same for get_fragment_testimonia_mentions.

src/rard/research/models/text_object_field.py
```diff
@@ -39,9 +41,40 @@ def get_related_object(self):
                 pass
         return None
 
+    fragment_testimonia_mentions = []
```
tcouch: I don't think storing mentions as a list attached to the model will work, because what happens to that list when the object is deleted from memory? It's not stored in the database, because it's not a field. You can get the same information by querying the reverse relationships on the m2m fields from fragments, testimonia and anonymous_fragments

acholyn:
Have implemented changes on the PR that pulls into this one since there were other issues fixed on there too

src/rard/research/tests/models/test_text_object_field.py
tcouch, line 1: Could do with a test for removing mentions

src/rard/research/forms.py
acholyn: ah, maybe it's an admin-only thing then. I'll remove it. I didn't see it show up in the form so we might not need a replacement either

PR #420, opened 2024-02-06
move links between works proposed changes

tcouch:
I was looking for a way to remove the none-option from the select boxes and went down a bit of a rabbit hole.
Anyway, I hope it's useful.

* Removes none-option from Work and Book select boxes when using form to update existing links
* Refactor the form's __init__ method to hopefully make it a bit more readable
* Update disable checkbox js to disable definite book as well as def work checkboxes when Unknown Work selected


acholyn (approved):
looks nice and works as expected so all good from me

PR #421, opened 2024-02-07
Feature/track ft mentions tc updates

tcouch:
I had a go at removing the fragment_testimonia_mentions list and this seems to work

### DIFF
--- a/src/rard/research/models/text_object_field.py
+++ b/src/rard/research/models/text_object_field.py
@@ -1,5 +1,6 @@
 from django.contrib.contenttypes.fields import GenericRelation
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import IntegrityError
 from simple_history.models import HistoricalRecords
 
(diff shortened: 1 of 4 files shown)


src/rard/research/models/text_object_field.py
```diff
@@ -41,40 +42,33 @@ def get_related_object(self):
...
+                try:
+                    class_mentions_queryset.add(pk)
+                except IntegrityError:
+                    pass
 
     def save(self, *args, **kwargs):
         # Update links generated from mentions each time we save
         self.link_bibliography_mentions_in_content()
-        self.update_mentions()
+        if not self._state.adding:
```
acholyn, line 70: is this a built in thing?

acholyn (approved):

src/rard/research/models/text_object_field.py
tcouch: Yeah, see here: https://docs.djangoproject.com/en/5.0/ref/models/instances/#state

PR #424, opened 2024-02-08
updating mentions delete behaviour so it cannot delete

### LINKED ISSUE #358: When deleting a fragment, alert users if there are @mentions referencing it

acholyn:
closes #358 
---
Updated behaviour so it's not possible to delete if there are mentions.


tcouch (approved):
I approve

PR #425, opened 2024-02-12
Fixing paste behaviour into text editor

### LINKED ISSUE #423: Issues with pasting into text editor triggering footnote section via carriage returns
When researchers paste from outside the editor it triggers the footnote divider, no content. It can also create a scrollable area. It doesn't paste the text at all, sometimes.

Based on this, it seems the Quill editor is mistakenly creating the footnotes area with a load of whitespace and this can't be removed since there are no footnotes. Researchers would like to keep the formatting of what's pasted so it wouldn't be ideal to force plain text pasting

acholyn:
closes #423 
---
The text editor was interpreting something (probably a carriage return) as a `<br>` which then got turned into a `<h4><br></h4>` and trigger the footnote area to be made with no content aside from the `<hr>`. This proved difficult to remove unless you knew where the element was but regardless, this behaviour wasn't expected. This can be bypassed on some browsers with the "force paste" option
The solution listens for a paste event and on paste removes the h4 br element inserted. This seems to work across browsers in my testing and doesn't seem to interfere with preexisting footnotes.

### DIFF
--- a/src/rard/static/js/project.js
+++ b/src/rard/static/js/project.js
@@ -241,37 +241,7 @@ async function getApparatusCriticusLines(searchTerm, object_id, object_class) {
   });
   return matches;
 }
-function textnode(node, delta) {
-  if (node.data) {
-    // console.log(
-    //   "textnode: node:",
-    //   node,
-    //   "node data",
-    //   node.nodeType,
-    //   node.nodeValue,
-    //   "delta",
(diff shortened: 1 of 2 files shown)


tcouch (approved):
Looks good. I couldn't find an example that'd replicate the original bug, but certainly seems to be working now anyway.

PR #426, opened 2024-02-14
Fix/412

### LINKED ISSUE #412: New bibliography tools
@acholyn. Hi Amanda, a couple of issues I've noticed when using the new bibliography tools:

There doesn't seem to be an ability to italicise text or use the other text-box features in the 'title' box of the new bibliography editor at the point when entries are added (correct formatting can currently only be implemented using the edit function, which isn't very obvious and takes several more clicks).

After deleting an item of bibliography added in error, the page displays incorrectly:

image

acholyn:
closes #412
---
Assumes it's only the title field that needs the editor. Have tested and seems to work fine; no issues using the previous field + rtf include

### DIFF
--- a/src/rard/templates/research/partials/rich_text_editor.html
+++ b/src/rard/templates/research/partials/rich_text_editor.html
@@ -2,7 +2,7 @@
 {# assumes form field as a context variable named 'field' and whether to enable mentions 'enable_mentions' and optional min_height and padding_bottom in pixels #}
 
 <div class='row mb-3 mt-n3 editor-area'>
-    <div class='col h-fill' style='min-height: {% if min_height %}{{ min_height }}{% else %}400{% endif %}px; padding-bottom: {% if padding_bottom != None %}{{ padding_bottom }}{% else %}60{% endif %}px;'>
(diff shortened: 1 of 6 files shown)


src/rard/templates/research/partials/rich_text_editor.html
```diff
@@ -2,7 +2,7 @@
 {# assumes form field as a context variable named 'field' and whether to enable mentions 'enable_mentions' and optional min_height and padding_bottom in pixels #}
 
 <div class='row mb-3 mt-n3 editor-area'>
-    <div class='col h-fill' style='min-height: {% if min_height %}{{ min_height }}{% else %}400{% endif %}px; padding-bottom: {% if padding_bottom != None %}{{ padding_bottom }}{% else %}60{% endif %}px;'>
+    <div class='col h-fill {% if extra_classes %}{{extra_classes}}{% endif %}' style='min-height: {% if min_height %}{{ min_height }}{% else %}400{% endif %}px; max-height: {% if max_height %}{{ max_height }}px{% else %}auto{% endif %}; padding-bottom: {% if padding_bottom != None %}{{ padding_bottom }}{% else %}60{% endif %}px;'>
```
tcouch, line 5: Just wondering at what point it's better to put the min-height, max-height, padding stuff in css and just apply a class?

tcouch (approved):
Looks good.

src/rard/templates/research/partials/rich_text_editor.html
acholyn: Like just apply a classname? That would be better practice, I think. I can open an issue for it to be done at some point

PR #427, opened 2024-02-16
Feature/duplicate fragment

### LINKED ISSUE #393: Duplicating fragments
Researchers would like the ability to duplicate a fragment with its original text (copy) and apparatus criticus but not the links. This would be a button available on the fragment page and could also be shown with the option of converting an Anon Frag to a normal one

acholyn:
closes #393 
---
New view with button to duplicate the fragment. The duplication also duplicates the `OriginalText` and the `Concordance` instances associated, as well as `Topics`, `References`, `Translations` and `ApparatusCriticusItems`. 

The code loops through all the fields of the `OriginalText` and copies their values and goes through these associated models and does the same thing, then assigns the new `OriginalText` instance in place of the original one. Finally, the `Fragment` is copied and the new `OriginalText` attached, and then `Topics` are copied.

On the frontend, the `Duplicate` button is clicked, a request is sent to the view with the pk of the original item. Upon completion, the user is redirected to the new unlinked fragment.

The commentary and links/apposita are not copied over, intentionally.

This has been implemented for `Fragments and `AnonymousFragments`, unlinked `Fragments` are created in both cases (tbc)

### DIFF
--- a/src/rard/research/tests/views/test_fragment.py
+++ b/src/rard/research/tests/views/test_fragment.py
@@ -175,7 +175,9 @@ class TestFragmentSuccessUrls(TestCase):
         # Simulate the behavior of the duplicate_fragment view
         request = RequestFactory().get("/")
         request.user = UserFactory.create()
-        response = duplicate_fragment(request, original_fragment.pk)
+        response = duplicate_fragment(
+            request, original_fragment.pk, model_name="fragment"
+        )
         duplicate_pk = response.url.split("/")[-2]
 
         expected_url = reverse(
--- a/src/rard/research/tests/views/test_fragment.py
+++ b/src/rard/research/tests/views/test_fragment.py
@@ -180,9 +180,7 @@ class TestFragmentSuccessUrls(TestCase):
         )
         duplicate_pk = response.url.split("/")[-2]
 
-        expected_url = reverse(
-            "fragment:detail", kwargs={"pk": duplicate_pk, "model_name": "fragment"}
-        )
+        expected_url = reverse("fragment:detail", kwargs={"pk": duplicate_pk})
 
         self.assertEqual(response.url, expected_url)
         self.assertEqual(response.status_code, 302)
--- a/src/rard/research/tests/views/test_fragment.py
+++ b/src/rard/research/tests/views/test_fragment.py
@@ -12,9 +12,12 @@ from rard.research.models import (
     CitingWork,
     Fragment,
     OriginalText,
+    Reference,
     Topic,
 )
 from rard.research.models.base import AppositumFragmentLink, FragmentLink
+from rard.research.models.linkable import ApparatusCriticusItem
+from rard.research.models.original_text import Concordance
 from rard.research.models.text_object_field import TextObjectField
 from rard.research.views import (
     AnonymousFragmentConvertToFragmentView,
@@ -27,6 +30,7 @@ from rard.research.views import (
     MoveAnonymousTopicLinkView,
     UnlinkedFragmentConvertToAnonymousView,
 )
+from rard.research.views.fragment import duplicate_fragment
 from rard.users.tests.factories import UserFactory
 from rard.utils.convertors import (
     FragmentIsNotConvertible,
@@ -163,6 +167,20 @@ class TestFragmentSuccessUrls(TestCase):
             view.get_success_url(),
             reverse("fragment:detail", kwargs={"pk": view.object.pk}),
         )
+
+    def test_duplication_success_url(self):
+        original_fragment = Fragment.objects.create(name="some name")
+
+        # Simulate the behavior of the duplicate_fragment view
+        request = RequestFactory().get("/")
+        request.user = UserFactory.create()
+        response = duplicate_fragment(request, original_fragment.pk)
+        duplicate_pk = response.url.split("/")[-2]
+
+        expected_url = reverse("fragment:detail", kwargs={"pk": duplicate_pk})
+
+        self.assertEqual(response.url, expected_url)
+        self.assertEqual(response.status_code, 302)
 
 
 class TestFragmentDeleteView(TestCase):
@@ -492,3 +510,48 @@ class TestOrderAnonymousFragmentListView(TestCase):
             response.context_data["object_list"],
             [self.aftl1, self.aftl3, self.aftl2, self.aftl3],
         )
+
+
+class TestFragmentDuplicationView(TestCase):
+    def setUp(self):
+        self.cw = CitingWork.objects.create(title="citing work title")
+        self.ot = OriginalText.objects.create(
+            content="Original Text test",
+            apparatus_criticus_blank=False,
+            apparatus_criticus="here be the app crit",
+            object_id=111,
+            citing_work=self.cw,
+            content_type_id=12,
+            reference_order="reference order",
+        )
+        self.c = Concordance.objects.create(
+            original_text=self.ot, source="tester", identifier="123"
+        )
+        self.ac = ApparatusCriticusItem.objects.create(
+            parent=self.ot, content="critical test", object_id=23
+        )
+        self.r = Reference.objects.create(editor="test", original_text=self.ot)
+        self.t = Topic.objects.create(name="topic1")
+        self.f = Fragment.objects.create(name="test fragment")
+        self.f.topics.add(self.t)
+        self.f.original_texts.add(self.ot)
+
+    def test_duplication(self):
+        url = reverse("fragment:duplicate", kwargs={"pk": self.f.pk})
+        request = RequestFactory().get(url)
+        request.user = UserFactory.create()
+        response = duplicate_fragment(request, pk=self.f.pk)
+        duplicate_pk = response.url.split("/")[-2]
+        duplicate_frag = Fragment.objects.get(pk=duplicate_pk)
+
+        for field in duplicate_frag._meta.fields:
+            if field.name in [
+                "id",
+                "created",
+                "modified",
+            ]:
+                continue
+            print(field.name)
+            value1 = getattr(duplicate_frag, field.name)
+            value2 = getattr(self.f, field.name)
+            self.assertEqual(value1, value2)
--- a/src/rard/research/tests/views/test_fragment.py
+++ b/src/rard/research/tests/views/test_fragment.py
@@ -14,6 +14,7 @@ from rard.research.models import (
     OriginalText,
     Reference,
     Topic,
+    Translation,
 )
 from rard.research.models.base import AppositumFragmentLink, FragmentLink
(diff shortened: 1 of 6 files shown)


src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+def duplicate_fragment(request, pk):
+    from django.contrib.contenttypes.models import ContentType
+
+    from rard.research.models import (
+        ApparatusCriticusItem,
+        Concordance,
+        OriginalText,
+        Reference,
+        Translation,
+    )
```
tcouch: You should only put import statements inside a function if you **really** have to (i.e. to avoid circular imports), as it's better to know if there's an import error when the module loads rather than when the function is called. Also it's easier to read all the imports in one place.

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+
+    # Get the original fragment
+
+    if "anonymous" in request.path:
```
tcouch: This approach feels a bit shaky to me - if we decided to change the url path for anonymous fragments for example, we'd have no way of knowing it'd break this function. See here for how to pass extra arguments to view functions as an alternative: https://docs.djangoproject.com/en/5.0/topics/http/urls/#passing-extra-options-to-view-functions

src/rard/research/urls.py
```diff
@@ -296,6 +296,7 @@
                         views.UnlinkedFragmentConvertToAnonymousView.as_view(),
                         name="convert_to_anonymous",
                     ),
+                    path("<pk>/duplicate", views.duplicate_fragment, name="duplicate"),
```
tcouch:
```suggestion
                    path("<pk>/duplicate", views.duplicate_fragment, {"model_name":"fragment"}, name="duplicate"),
```

src/rard/research/urls.py
```diff
@@ -368,6 +369,7 @@
                         views.AnonymousFragmentDeleteView.as_view(),
                         name="delete",
                     ),
+                    path("<pk>/duplicate", views.duplicate_fragment, name="duplicate"),
```
tcouch:
```suggestion
                    path("<pk>/duplicate", views.duplicate_fragment, {"model_name":"anonymousfragment"}, name="duplicate"),
```

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
         "research/partials/htmx_fragment_results.html",
         {"fragments": fragments},
     )
+
+
+def duplicate_fragment(request, pk):
```
tcouch:
```suggestion
def duplicate_fragment(request, pk, model_name):
```

src/rard/research/views/fragment.py
tcouch: So something like:
```
if model_name == "anonymousfragment":
    og_fr = ...
elif model_name == "fragment":
    og_fr = ...
else:
    raise Error("model name not recognised")
```

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+
+    if "anonymous" in request.path:
+        original_fragment = get_object_or_404(AnonymousFragment, pk=pk)
+    else:
+        original_fragment = get_object_or_404(Fragment, pk=pk)
+    original_original_texts = original_fragment.original_texts.all()
+
+    new_original_texts = []
+
+    for i in range(len(original_original_texts)):
```
tcouch: Why iterate over a range when you could just do: `for text in original_original_texts:` ?

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+        for field in original_original_texts[i]._meta.fields:
+            # Exclude the ID field
+            if field.name in ["id", "created", "modified"]:
+                continue
+
+            field_value = getattr(original_original_texts[i], field.name)
+
+            # For ForeignKey fields, copy the related object
+            if field.is_relation and field_value is not None:
+                field_value = field_value
```
tcouch: I don't think this changes anything?

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+        # Copy relationships to Concordances, Ap Crit & Translations
+        model_details = {
+            Concordance: {
+                "filter_name": "original_text",
+                "filter_value": original_original_texts[i],
+                "ot_fieldname": "original_text",
+            },
+            ApparatusCriticusItem: {
+                "filter_name": "object_id",
+                "filter_value": original_original_texts[i].pk,
```
tcouch: Generic relation on original text model allows this filter to work
```suggestion
                "filter_name": "original_text",
                "filter_value": original_original_texts[i],
```

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+            ):
+                for field in original_item._meta.fields:
+                    # Exclude some fields
+                    if field.name in [
+                        "id",
+                        "created",
+                        "modified",
+                        ot_fieldname,
+                        "content_type",
+                    ]:
```
tcouch, line 1098: You can also add object_id to this list as it and content type are inferred from parent.
```suggestion
                        "object_id",
                    ]:
```

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+                new_model_data[
+                    ot_fieldname
+                ] = new_original_text  # make sure to assign new OT in place of the old one (field skipped above)
+
+                new_model = model_class.objects.create(**new_model_data)
+                # specific handling of content_type since we don't want to create a new instance
+                if hasattr(model_class, "content_type") and model_class.content_type:
+                    new_model.content_type = ContentType.objects.get(
+                        pk=original_item.content_type.pk
+                    )
```
tcouch: I don't understand what this bit is for. When you create an AppCritItem with parent=new_original_text, the content type is inferred from the class of the object you pass as the parent

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+
+    # Create a new fragment with the same values as original
+    new_fragment_data = {}
+    for field in original_fragment._meta.fields:
+        # Exclude unique fields
+        if field.name in [
+            "id",
+            "created",
+            "modified",
+            "commentary",
```
tcouch, line 1118: `plain_commentary` too?

src/rard/research/views/fragment.py
```diff
@@ -1010,3 +1010,136 @@ def fetch_fragments(request):
...
+            "order",  # anon frags have this
+            "model",
```
tcouch: `model` - is that a field? I can't see it.

src/rard/templates/research/anonymousfragment_detail.html
```diff
@@ -41,6 +41,13 @@
                     >{% trans 'Make non-anonymous' %}</button>
                 </div>
             </form>
+            <form novalidate style='margin-left:auto' class='form-inline' action='{% url "anonymous_fragment:duplicate" object.pk %}' method='POST'>
+                {% csrf_token %}
+                <div class='form-group'>
+                    <button type='submit' class='btn btn-link text-info p-0 ml-2'
+                        data-what='fragment'>{% trans 'Duplicate' %}</button>
```
tcouch: Just for consistency with above
```suggestion
                        data-what='anonymous fragment'>{% trans 'Duplicate' %}</button>
```

src/rard/research/tests/views/test_fragment.py
```diff
@@ -12,9 +12,13 @@
     CitingWork,
     Fragment,
     OriginalText,
+    Reference,
     Topic,
+    Translation,
 )
 from rard.research.models.base import AppositumFragmentLink, FragmentLink
+from rard.research.models.linkable import ApparatusCriticusItem
+from rard.research.models.original_text import Concordance
```
tcouch: Can both be imported directly from rard.research.models

src/rard/research/tests/views/test_fragment.py
```diff
@@ -27,6 +31,7 @@
     MoveAnonymousTopicLinkView,
     UnlinkedFragmentConvertToAnonymousView,
 )
+from rard.research.views.fragment import duplicate_fragment
```
tcouch: Can be imported directly from views

src/rard/research/tests/views/test_fragment.py
```diff
@@ -492,3 +511,131 @@ def test_compare_orders(self):
...
+        self.cw = CitingWork.objects.create(title="citing work title")
+        self.ot = OriginalText.objects.create(
+            content="Original Text test",
+            apparatus_criticus_blank=False,
+            apparatus_criticus="here be the app crit",
+            object_id=111,
+            citing_work=self.cw,
+            content_type_id=12,
+            reference_order="reference order",
+        )
```
tcouch: apparatus_criticus isn't used, object_id and content_type (rather than content_type_id) shouldn't be set directly (I don't think), instead create the fragment first and set owner = fragment.

src/rard/research/tests/views/test_fragment.py
```diff
@@ -492,3 +511,131 @@ def test_compare_orders(self):
...
+                "created",
+                "modified",
+                "commentary",
+                "object_id",
+                "original_text",
+                "order",
+                "model",
+            ]:
+                continue
+            if field.is_relation and getattr(original, field.name) is not None:
```
tcouch: Why not even if it is none?

tcouch (changes requested):

src/rard/research/views/fragment.py
acholyn: so it should work without this but also having it in the list above so it's ignored?

src/rard/research/views/fragment.py
acholyn: it would distinguish between af and frag but I'll try without it and see if it throws the error it did before

src/rard/research/tests/views/test_fragment.py
acholyn: think it threw an error but may have just been copied over

src/rard/research/views/fragment.py
acholyn: image
It may only be applicable to the test since on the frontend it will still go through the process but I thought it was better to keep in line with tests?

tcouch (approved):

PR #428, opened 2024-02-20
Feature/commentary mentions box

### LINKED ISSUE #413: Box for @ mentions
We have been told that commentaries are not going to be kept in the public version.

We would like to have a separate box below the apparatus criticus just for @ mentions.

acholyn:
closes #413 
---
Created new `PublicCommentaryMentions` model, which is a simpler version of `TextObjectField`, it has a 1:1 relationship with models that have commentaries. On the detail page, this section can be toggled when editing and can be edited itself. It does not have a plain version, nor tracks mentions as it should be a slimmed down version of the actual `Commentary`, purely for public consumption and ergo the duplication is unnecessary.

It utilises the same htmx inline update flow as introductions and commentaries.


src/rard/templates/research/fragment_detail.html
```diff
@@ -179,16 +179,56 @@ <h5>{% trans 'Commentary' %}</h5>
...
+    <section class="public-commentary-section" style="display:none;">
+        <div class='d-flex justify-content-between mb-3'>
+          <div>
+              <h5>{% trans 'See Also' %}</h5>
+          </div>
+          <div>
+              {% if perms.research.change_fragment and has_object_lock %}
+              <button
+                class="btn text-primary p-0 ml-2"
+                hx-get="{% url 'anonymous_fragment:update_public_commentary' object.pk %}"
```
tcouch:
```suggestion
                hx-get="{% url 'fragment:update_public_commentary' object.pk %}"
```

src/rard/templates/research/anonymousfragment_detail.html
```diff
@@ -166,17 +167,58 @@ <h5>{% trans 'Commentary' %}</h5>
...
 
 
+    </section>
+    {% if perms.research.change_fragment and has_object_lock %}
+    <br>
+    <div class="custom-control custom-switch">
+        <input type="checkbox" class="custom-control-input" id="togglePublicCommentary">
+        <label class="custom-control-label text-muted" for="togglePublicCommentary">Show Public Commentary</label>
+      </div>
+    <section class="public-commentary-section" style="display:none;">
```
tcouch: Please try to make sure opening and closing tags have the same indentation

src/rard/templates/research/anonymousfragment_detail.html
```diff
@@ -166,17 +167,58 @@ <h5>{% trans 'Commentary' %}</h5>
...
+              {% if perms.research.change_fragment and has_object_lock %}
+              <button
+                class="btn text-primary p-0 ml-2"
+                hx-get="{% url 'anonymous_fragment:update_public_commentary' object.pk %}"
+                hx-target=".public-commentary-container"
+                hx-swap="innerHTML">{% trans 'Edit' %}</button>
+              {% endif %}
+          </div>
+      </div>
+      {% if object.public_commentary_mentions.approved %}
```
tcouch: Can't edit public commentary because .public-commentary-container only included if approved (default is false)

tcouch:
One small bug:
* Create a public commentary and tick the box to say it's approved
* Finish editing and go back to the fragment detail page - you should see both public and private commentaries
* Now click to edit again and you'll see the public commentary displayed, but the toggle to show public commentary is off
* Click the toggle and now the public commentary is shown twice

And another (probably my fault from when I set it up)
* The cancel button doesn't work on the inline main commentary update form

Do we actually need the toggle on the edit page?

acholyn:
we might not even need the toggle in that case, we could show it in all cases except when not approved and not edit (will show for not approved and edit, approved and edit & approved and not edit)

tcouch:
Ah, one more thing:

1. Edit a Testimonium
2. Create public commentary
3. Do not tick approved
4. Finish editing
5. The detail page still shows the public commentary, but "See Also" isn't shown

Not sure if it's just Tesimonia or all types

acholyn:
Weird, that's not the behaviour on my machine - if it's not approved it doesn't show. Did you pull in the latest changes?
When it has been approved, the second box will show without the 'see also' underneath the commentary; not sure if in the final version they want to just leave it being called commentary or not

acholyn:
ah, ooh I've got a new bug where it shows up twice though

PR #430, opened 2024-02-23
Feature/maintain mentions conversion

### LINKED ISSUE #296: When converting unlinked to anonymous or reverse maintain @mentions
At the moment when the conversion happens, the original fragment is deleted. We need to fix any broken @mentions in commentaries etc.

acholyn:
closes #296 
---
Conversion from `unlinked` to `anonymous` and the reverse should now transfer the mentions by updating the `TOF` in which they are mentioned by changing the `data` parameters, followed by the `update_content_mentions` which will update the display text. The `update_mentions()` function should handle the updating of the `mentioned_in` for the new fragment.


src/rard/utils/basemodel.py
```diff
@@ -41,12 +41,31 @@ def get_fragment_testimonia_mentions(self):
...
-                    model = item.attrs.get("data-target", None)
+                    model = item.attrs.get("data-target", None).lower()
                     pk = item.attrs.get("data-id", None)
                     if model in linked_items.keys() and pk:
                         linked_items[model].append(pk)
                 return linked_items
 
+            def reassign_mentions(self, original, new):
+                value = getattr(self, field_name)
+                soup = soup = bs4.BeautifulSoup(value, features="html.parser")
```
tcouch: double soup
```suggestion
                soup = bs4.BeautifulSoup(value, features="html.parser")
```

src/rard/research/models/text_object_field.py
```diff
@@ -51,6 +53,12 @@ def update_mentions(self):
             # get related queryset
             class_mentions_queryset = getattr(self, class_name + "_mentions")
             for instance in class_mentions_queryset.all():
+                try:
+                    get_object_or_404(instance.__class__, pk=instance.pk)
+                except Http404:
+                    instance.mentioned_in.remove(
+                        self
+                    )  # if instance no longer exists, remove it from mentioned_in
```
tcouch: A queryset wouldn't contain objects that don't exist. I don't understand why this would be necessary. You've got a reference to an instance and then using that instance's properties to look it up, so how would that ever result in it not being found?

src/rard/utils/convertors.py
```diff
@@ -31,6 +31,17 @@ def transfer_data_between_fragments(source, destination):
     destination.commentary = commentary
 
 
+def transfer_mentions(original, new):
+    if original.mentioned_in:
+        for tof in original.mentioned_in.all():
+            tof.reassign_mentions(original, new)
+            tof.update_content_mentions()
+            tof.update_mentions()
```
tcouch: Can you add a comment next to each function call to summarise what it's doing?
e.g. reassigning mentions in the tof content, updating mention display text, updating database relations...

tcouch (approved):

src/rard/research/models/text_object_field.py
acholyn: if an object that's been deleted is mentioned, I thought it's still in the content and only in the non editing view (TOF preview) is it changed to `bad-link`? I ran into this issue previously but possibly the full flow fixes that?

acholyn: ah - calling the `update_content_mentions` should handle that so it shouldn't be an issue anymore if they manage to delete it but otherwise they can't delete it because it's been mentioned elsewhere

Issue #362, opened 2022-09-22
Add search substitutions for greek characters with diacritics


tcouch:
Currently the exact greek character including diacritics needs to be present in search terms in order to match content. Can we use search substitutions to make this more flexible?

tcouch:
Will need help from a member of the research team on this one to understand the set of substitutions we'd need.

tcouch:
Possibly related to #64

tcouch:
Contact @dgiordani when ready to start

PR #431, opened 2024-02-23
Feature/search diacriticals

### LINKED ISSUE #377: Searching Greek with diacriticals
Searched of Greek text must employ the correct diacriticals to obtain results (e.g. διὰ not δια). Ideally, a search for δια should also return διὰ. 
Could this be implemented using symbol substitution, like the current search functionality for Latin text that use u/v, etc.?

acholyn:
closes #377 and #362 
---
Plain text versions of text objects have the unicode characters stripped out (non ascii) so a text search of "plain text" should return results matching "Õh Ṭọ hάvë pļāĭñ těxt" as the diacriticals etc are ignored.


src/rard/research/tests/views/test_search.py
```diff
@@ -548,3 +548,30 @@ def do_search(search_function, keywords):
...
+        f3 = Fragment.objects.create()
+        f3.commentary = TextObjectField.objects.create(content="Ĺĭōṇęḷ ώὰs Ᾰ gΌῸd ὲgg")
+
+        ot3 = OriginalText.objects.create(
+            content="Õh Ṭọ hάvë pļāĭñ těxt",
+            citing_work=cw,
+            owner=f3,
+        )
+
+        Translation.objects.create(original_text=ot3, translated_text="Ὼῤῥῠ")
```
tcouch: You haven't used this

tcouch (changes requested):
I just realised unidecode converts everything to ascii which means searches for greek characters wouldn't work. Either we'd have to convert search terms to ascii as well, or use another method such as this one to remove diacritics:

```
def remove_combining_fluent(string: str) -> str:
    normalized = unicodedata.normalize('NFD', string)
    return ''.join(
        [l for l in normalized if not unicodedata.combining(l)]
    ).casefold()
```

Also, we need to write a migration to recreate plain content for all the relevant text fields using the new approach.

acholyn:
The searches for the Greek characters are converted to their translated version eg. α is a , from what I could tell, is that not helpful? Do they enter Greek characters in the search?

src/rard/research/tests/views/test_search.py
acholyn: doesn't seem to work when using the fragment search, should it not?

tcouch: "Ὼῤῥῠ" should get converted to "ORRu" - is that what you searched for?

tcouch:
Richard's original comment in #377 mentions using Greek characters in the search. Like I say, we could solve it by converting the search term using unidecode as well, but I feel like the function I suggested above which just strips out diacritics without converting to ASCII would be safer.

src/rard/research/tests/views/test_search.py
acholyn: searched for "OR"

acholyn:
> I just realised unidecode converts everything to ascii which means searches for greek characters wouldn't work. Either we'd have to convert search terms to ascii as well, or use another method such as this one to remove diacritics:
> 
> ```
> def remove_combining_fluent(string: str) -> str:
>     normalized = unicodedata.normalize('NFD', string)
>     return ''.join(
>         [l for l in normalized if not unicodedata.combining(l)]
>     ).casefold()
> ```
> 
> Also, we need to write a migration to recreate plain content for all the relevant text fields using the new approach.

what does the NFD mean in this code?

tcouch:
> > I just realised unidecode converts everything to ascii which means searches for greek characters wouldn't work. Either we'd have to convert search terms to ascii as well, or use another method such as this one to remove diacritics:
> > ```
> > def remove_combining_fluent(string: str) -> str:
> >     normalized = unicodedata.normalize('NFD', string)
> >     return ''.join(
> >         [l for l in normalized if not unicodedata.combining(l)]
> >     ).casefold()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Also, we need to write a migration to recreate plain content for all the relevant text fields using the new approach.
> 
> what does the NFD mean in this code?

Normal Form D. From the documentation for [unicodedata](https://docs.python.org/3/library/unicodedata.html#unicodedata.normalize):

> ...there are two normal forms: normal form C and normal form D. Normal form D (NFD) is also known as canonical decomposition, and translates each character into its decomposed form. Normal form C (NFC) first applies a canonical decomposition, then composes pre-combined characters again.

So, it splits a character with a diacritic into two parts. The next part of that function then discards any parts that are combining characters.

src/rard/utils/text_processors.py
```diff
@@ -1,11 +1,23 @@
...
 
 
+def strip_combining(content):
+    """Converts the content to their base and combining characters,
+    then removes the combining ones and returns a lowercase string of the base characters
+    """
+    normalized = unicodedata.normalize("NFD", content)
+    return "".join(
+        [char for char in normalized if not unicodedata.combining(char)]
+    ).casefold()
```
tcouch: I think casefold is the cause of our failing test. Probably best to just remove it.

tcouch:
```suggestion
    )
```

src/rard/research/migrations/0070_updating_plain_renditions.py
```diff
@@ -0,0 +1,100 @@
...
+from django.db import migrations, models
+from django.utils.html import strip_tags
+
+
+def strip_combining_and_make_plain(content):
+    plain_text = make_plain_text(content)
+    normalized = unicodedata.normalize("NFD", plain_text)
+    return "".join(
+        [char for char in normalized if not unicodedata.combining(char)]
+    ).casefold()
```
tcouch:
```suggestion
    )
```

src/rard/research/migrations/0070_updating_plain_renditions.py
```diff
@@ -0,0 +1,100 @@
...
+    no_lone_numbers = re.sub(r"\s\d{1,2}\s", " ", no_punctuation)  # mentions
+    no_excess_space = re.sub(r" +", " ", no_lone_numbers)
+    return no_excess_space
+
+
+def update_antiquarians(apps, schema_editor, plaining_function):
+    Antiquarian = apps.get_model("research", "Antiquarian")
+    for ant in Antiquarian.objects.all():
+        if ant.plain_introduction:
+            plaining_function(ant.introduction.content)
```
tcouch: Both plaining functions return a string which you're not doing anything with. You need to assign the result to `ant.plain_introduction`. Same for all the other models

src/rard/research/migrations/0070_updating_plain_renditions.py
```diff
@@ -0,0 +1,100 @@
...
+    OriginalText = apps.get_model("research", "OriginalText")
+    for ot in OriginalText.objects.all():
+        if ot.plain_content:
+            plaining_function(ot.content)
+        if ot.plain_translated_text:
+            plaining_function(ot.translated_text)
+        ot.save()
+
+
+def update_plain_texts(apps, schema_editor):
```
tcouch, line 76: A generic form of this could be something like:
```
PLAIN_TEXT_MODEL_FIELDS = {
    "Antiquarian": "introduction",
    "Fragment":"commentary",
    "OriginalText":"content",
    "Translation":"translated_text"
    ...etc
   }
 
def update_plain_text(apps, model_name, field_name, plaining_function):
    model = apps.get_model("research", model_name)
    for instance in model.objects.all():
        content = getattr(instance, field_name)
        setattr(instance, "plain_" + field_name, plaining_function(content))
        instance.save()

def update_plain_texts(apps, schema_editor):
    for model_name, field_name in PLAIN_TEXT_MODEL_FIELDS.items():
        update_plain_text(apps, model_name, field_name, strip_combining_and_make_plain)
```

tcouch (changes requested):

src/rard/research/migrations/0070_updating_plain_renditions.py
```diff
@@ -25,70 +23,33 @@ def make_plain_text(content):
...
-    for ot in OriginalText.objects.all():
-        if ot.plain_content:
-            ot.plain_content = plaining_function(ot.content)
-        if ot.plain_translated_text:
-            ot.plain_translated_text = plaining_function(ot.translated_text)
-        ot.save()
+def update_plain_text(apps, model_name, field_name, plaining_function):
+    model = apps.get_model("research", model_name)
+    for instance in model.objects.all():
+        content = getattr(instance, field_name)
```
tcouch: Really sorry I think this isn't as straightforward as that because for Fragments it's `instance.commentary.content`. Probably best to revert this for the longer version, which we know works.

tcouch (approved):

PR #436, opened 2024-02-28
Feature/shorten mentions

### LINKED ISSUE #322: Shorten information displayed in bibliography @mentions
Currently, the full author name and title is provided with each mention, but all we need is the author surname and the date of publication in brackets, with no following space or punctuation, i.e. Chappuis (1868), not Chappuis [1868]: Fragments des ouvrages de M. Terentius Varron intitulés Logistorici, Hebdomades vel De imaginibus, De forma philosophiae , Paris. (very high priority)

acholyn:
closes #322 
---
- adds a new function to `BibliographyItem`s which gets passed in the mentions search
- mentions dropdown should display full version but the inserted version is the citation, which is what's persisted once in the editor (meaning the same when the object is saved and opened again for editing as other functions often change this)

### DIFF
--- a/src/rard/research/views/mention.py
+++ b/src/rard/research/views/mention.py
@@ -178,7 +178,9 @@ class MentionSearchView(LoginRequiredMixin, View):
             if not model_name:
                 model_name = next(k for k, value in dd.items() if value == o.__class__)
                 model_name_cache[o.__class__] = model_name
-            citation = o.mention_citation() if hasattr(o,"mention_citation") else str(o)
+            citation = (
+                o.mention_citation() if hasattr(o, "mention_citation") else str(o)
+            )
             ajax_data.append(
                 {
                     "id": o.pk,
--- a/src/rard/research/views/mention.py
+++ b/src/rard/research/views/mention.py
@@ -178,19 +178,15 @@ class MentionSearchView(LoginRequiredMixin, View):
             if not model_name:
                 model_name = next(k for k, value in dd.items() if value == o.__class__)
                 model_name_cache[o.__class__] = model_name
-            if model_name == "bibliographyitem":
-                ajax_data.append(
-                    {
-                        "id": o.pk,
-                        "target": model_name,
-                        "value": str(o),
-                        "citation": BibliographyItem.objects.get(
-                            pk=o.pk
-                        ).mention_citation(),
-                    }
-                )
-            else:
-                ajax_data.append({"id": o.pk, "target": model_name, "value": str(o)})
+            citation = o.mention_citation() if hasattr(o,"mention_citation") else str(o)
+            ajax_data.append(
+                {
+                    "id": o.pk,
+                    "target": model_name,
+                    "value": str(o),
+                    "citation": citation,
+                }
+            )
         return JsonResponse(data=ajax_data, safe=False)
 
     def parse_mention(self, q):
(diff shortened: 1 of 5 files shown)


src/rard/research/views/mention.py
```diff
@@ -178,8 +178,19 @@ def get(self, request, *args, **kwargs):
...
+                        "id": o.pk,
+                        "target": model_name,
+                        "value": str(o),
+                        "citation": BibliographyItem.objects.get(
+                            pk=o.pk
+                        ).mention_citation(),
+                    }
+                )
+            else:
+                ajax_data.append({"id": o.pk, "target": model_name, "value": str(o)})
```
tcouch: A more resilient approach might be:
```suggestion
            citation = o.mention_citation() if hasattr(o,"mention_citation") else str(o)
            ajax_data.append(
                {
                    "id": o.pk,
                    "target": model_name,
                    "value": str(o),
                    "citation": citation,
                }
            )
```

tcouch: That'd leave you the flexibility to create custom citations for other models by adding a mention_citation method if the need arises

src/rard/static/js/project.js
```diff
@@ -405,10 +412,18 @@ function initRichTextEditor($item) {
         }
       },
       renderItem(item, searchTerm) {
-        // allows you to control how the item is displayed
+        // allows you to control how the item is displayed in the list
         let list_display = item.list_display || item.value;
         return `${list_display}`;
       },
+      onSelect(item, insertItem) {
+        if (item.citation) {
```
tcouch: If you apply the suggestion above, this can be simplified since all items should have a citation

src/rard/utils/basemodel.py
```diff
@@ -79,7 +79,10 @@ def update_editable_mentions(self, save=True):
                             linked = model.objects.get(pk=int(pkstr))
 
                             if char == "@":
-                                link_text = str(linked)
+                                if model_name == "bibliographyitem":
```
tcouch: Similar to above: `if hasattr(linked, "mention_citation")...`

src/rard/utils/basemodel.py
```diff
@@ -160,12 +163,21 @@ def render_dynamic_content(self):
                             linked = model.objects.get(pk=int(pkstr))
                             # is it something we can link to?
                             if getattr(linked, "get_absolute_url", False):
-                                replacement = bs4.BeautifulSoup(
-                                    '<a href="{}">{}</a>'.format(
-                                        linked.get_absolute_url(), str(linked)
-                                    ),
-                                    features="html.parser",
-                                )
+                                if model_name == "bibliographyitem":
```
tcouch: `if hasattr(linked, "mention_citation"):`

tcouch (changes requested):

src/rard/static/js/project.js
acholyn: I thought this was to distinguish eg antiquarians and bibliographies? Bibs have a citation, I didn't think antiquarians did, do they?

tcouch: With the change to views/mention.py above, the ajax_data always includes a citation for each item - it's just that for everything but Bibliographies it's the same as item.value

acholyn:
when searching: 
image

when entered: 
image
image

tcouch (approved):

PR #437, opened 2024-02-28
Fix/429

### LINKED ISSUE #429: abstract min/max height etc for rich text editor into classes and apply those instead
Better practice than applying styling directly to the element
Raised in #426

acholyn:
closes #429 
---
- have made a base class `rich-editor__wrapper` which utilises the defaults of the inline styling
- kept the `extra_classes` parameter for flexibility
- created classes for the instances where the default inline styling was overwritten and used them in `extra_classes`
- so long as any new classes are written in the CSS _after_ the `rich-editor__wrapper` they will take precedence in the attributes defined

### DIFF
--- a/src/rard/static/css/project.css
+++ b/src/rard/static/css/project.css
@@ -390,7 +390,7 @@ button.insert {
   background-color: #ffff00;
 }
 
-/* editor related */
+/****************** editor related *********************/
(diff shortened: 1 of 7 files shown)


tcouch (approved):
Bueno

PR #438, opened 2024-03-05
Feature/adding fraction symbols

### LINKED ISSUE #24: Create SVG files for non-standard symbols that are not included in the alphabetum font
Richard sent a file with some fraction symbols. Create those and add to symbol picker.
[Roman Fractions.pdf](https://github.com/UCL/frrant/files/7467185/Roman.Fractions.pdf)

See also http://scholarsfonts.net/claslat_13.pdf for other rare symbols that might need to be manually added to the Alphabetum font.

### LINKED ISSUE #387: Document how to add rare characters to Alphabetum font
Some characters are not available in the alphabetum font and there should be a way to add characters to the font. This needs to be documented

### LINKED ISSUE #448: Add new characters to the alphabetum font
as mentioned by @rmmarshall there are some more symbols to be added to the alphabetum font

acholyn:
closes #24 
---
Also closes #448 and #387 since the process is now known
- fraction symbols have been added, the Volusius Maecianus and Victorius each to their own symbol groups and the names correspond to the latin name provided in the table. If it would be better to rename them to their respective numerical fractions, that can be done through the admin
  - there was a little bit of overlap of the unicode codes with some of the Latin with vowel length characters, these were the same symbol so they've been deleted from this group. The other symbols in that group do not display on local and I can't see where they're imported from - as this conflict wasn't known before one of the groups will have to change the unicode code for 4 symbols: E00A, E010, E011, E019
- font styling has been abstracted back into the `alphabetum.css` file and this is imported into the base template
- the symbols were created with https://www.glyphrstudio.com/app/ and exported as an otf file. I will retain copies of the project files in case they need to be updated (particularly regarding the widths of characters)
- these symbols cannot be used with the vincula or underdot in the text editor


acholyn:
Have remapped the new custom ones:
E019 - victorius quadrans -> E023
E00A - VM sescuncia -> E024
E010 - VM dextans -> E026
E011 - VM deunx -> E027

tcouch:
There's a file `src/compose/local/django/symbols.json` that gets loaded into the database by the start script every time you bring the docker containers up. That will need updating by dumping the Symbol and SymbolGroup models into json once you've added the new symbols via the admin.

It looks like there's a separate symbols.json for each of local, development, and production though I'm not sure why (they're all the same size so I assume they're identical). They'll all need updating.

Also, there's a weird bug where custom symbols I've added to the symbol picker are really hard to click on. Is that just a firefox thing or do you have that problem too?

acholyn:
I remember there was a bit of strange behaviour with the bottom row of symbols not being selectable when they didn't have width but since fixing that I haven't had any issues

tcouch:
Also, at some point the alphabetum file name changed from `/static/fonts/ALPHABETUM_v_12.00.otf` to `/static/fonts/ALPHABETUM_v_12.otf` so the css file will need updating

acholyn:
image
image

tcouch (commented):
When starting with a blank database I get an error saying the symbolgroup with pk=17 doesn't exist. I think this is because you haven't included the new symbol groups in symbols.json?

```
 Traceback (most recent call last):
django         |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 82, in _execute
django         |     return self.cursor.execute(sql)
django         | psycopg2.errors.ForeignKeyViolation: insert or update on table "research_symbol" violates foreign key constraint "research_symbol_group_id_50891aa4_fk_research_symbolgroup_id"
django         | DETAIL:  Key (group_id)=(17) is not present in table "research_symbolgroup".
```

src/compose/production/django/symbols-pre-2024.json
tcouch, line 1: What's the purpose of these pre-2024 files?

acholyn: having a backup of the symbols before I added the custom ones - just in case something ends up being buggy.

tcouch:
This all looks good to me now. Just to say, if you're keeping a pre-2024 copy around in case you want to go back later, it's under version control so you can just checkout the old version of that file from a previous commit.

PR #439, opened 2024-03-08
Feature/duplicate fragments

### LINKED ISSUE #393: Duplicating fragments

acholyn:
closes #393 
---
- Have expanded the current duplication functionality to include tracking of duplicates (similarly to mentions) and alerting users when attempting to edit certain facets of an anonymous/fragment if there are duplicates.
- This functionality is extended to when fragments and anonymous fragments are converted between
- When duplicating an anonymous/fragment the new fragment is added to the duplicates field 
- There is a field for anonymous fragments to be used when a conversion has happened
- they have a property called duplicates_list which combines duplicates from the property and the reverse relationships


tcouch:
A couple of things still to do:
* Duplicating a duplicate should make the new fragment a duplicate of both the original and the first duplicate.
* Users should see a warning that a fragment has duplicates if they try to edit it. Maybe a banner at the top?

acholyn:
would it work in that when duplicating, if the original has duplicates, add those items to the new instance's list of duplicates and it to theirs?

.github/workflows/docker.yml
```diff
@@ -37,21 +37,22 @@ jobs:
...
-        run:  |
+        run: |
           cd src
           docker-compose -f local.yml build
 
       - name: Run DB Migrations
-        run:  |
+        run: |
           cd src
+          docker-compose -f local.yml run --rm django python manage.py makemigrations --merge --noinput
```
tcouch: What's this for?

src/rard/research/migrations/0070_add_duplicates_field.py
```diff
@@ -0,0 +1,67 @@
...
+            model_name="fragment",
+            name="duplicate_afs",
+            field=models.ManyToManyField(
+                blank=True,
+                related_name="fragment_duplicate_anonfragments",
+                to="research.AnonymousFragment",
+            ),
+        ),
+        migrations.AddField(
+            model_name="testimonium",
```
tcouch: Did we raise the issue of what to do about testimonia? I can see there's no option to duplicate testimonia on the detail page, but wasn't sure if we'd asked about this?

If it's not an option, these fields are redundant, but I guess not causing any harm. Maybe just add a comment here and in models/base.py to state that these are here and why if that's the case.

src/rard/research/views/fragment.py
```diff
@@ -1282,4 +1282,17 @@ def duplicate_fragment(request, pk, model_name):
...
+        for frag in original_fragment.duplicate_frags.all():
+            new_fragment.duplicate_frags.add(frag)
+        for af in original_fragment.duplicate_afs.all():
+            new_fragment.duplicate_afs.add(af)
+
+    original_fragment.duplicate_frags.add(new_fragment)
+    if model_name == "fragment":
+        new_fragment.duplicate_frags.add(original_fragment)
+    elif model_name == "anonymousfragment":
+        new_fragment.duplicate_afs.add(original_fragment)
```
tcouch, line 1296: I don't think you need this because adding new_fragment to original_fragment.duplicate_frags should add it in both directions. Only 80% sure on that though.

tcouch (changes requested):
I get an error when converting an anonymous fragment with duplicates into an unlinked fragment. To reproduce:
1. Duplicate an anonymous fragment
2. Convert the original anonymous fragment to unlinked

Result:

`ValueError: Cannot query "Unlinked 450": Must be "AnonymousFragment" instance.`

.github/workflows/docker.yml
acholyn: When we merged and the migrations were out of order but had technically also been applied already it was getting upset. Since it's run once do you think it'll have fixed the issue and we can remove it now?

src/rard/research/migrations/0070_add_duplicates_field.py
acholyn: I'm not sure we did ask if we should enable it for them, I'll make note to do that. They are currently redundant but are there because it's on the `HistoricalBaseModel` and it complained when I removed this from the migration.

src/rard/research/views/fragment.py
acholyn: I think technically you're correct but the way to access it is through a different sort of query and I found it simpler to do it this way, which also simplified the `duplicates_list` function

acholyn:
conversion issue should be fixed now but initially the new fragment page didn't load and I had to reload for it to display, wondering if this was specific to my device

.github/workflows/docker.yml
tcouch: It should just be building a fresh db every time so I'm not sure why it would have been a problem.

acholyn: perhaps there's another reason? I'm not really sure but [this](https://github.com/UCL/frrant/actions/runs/8452099681/job/23151772385) was the error output so maybe it caches the build within PRs (if that would explain the behaviour)

tcouch: Pretty sure that error means there's two migration files with the same earlier migration file listed as a dependency. Now that that's fixed I think it's best to remove it as it could mask any future problems.

src/rard/research/views/fragment.py
tcouch: Fair enough

tcouch (approved):

tcouch:
I think the migration leaf node error we're seeing is because GH actions actually runs some kind of merged version of this branch and development together. Usually it's quite good at telling you there's a conflict, but in the case of migrations it's not clever enough to warn that development has a 0070 migration file and so does this branch.

TL;DR: Merge development into this branch and resolve the migration conflict by making this branch's migration file 0071.

PR #441, opened 2024-03-12
Allow Anonymous Fragments to have apposita

### LINKED ISSUE #374: Adding apposita to anonymous fragments
@tcouch 
Hi Tom, would it be possible to enable us to add apposita to anonymous fragments? Now that we have a better understanding of the material in the database, this feature might be useful afterall.

tcouch:
closes #374 

In the Appositum Links section at the bottom of the Anonymous Fragment detail page, users can now choose to link another Anonymous Fragment. This will make the first AF an apposita of the AF they select.

These links are displayed in the Appositum Links section of the apposita, or in the Apposita section of the AF being linked to. An unlink button in the Appositum Links list allows these links to be deleted providing the apposita has a lock.

Also created unlink option for unlinked fragments' apposita as this wasn't previously possible.


tcouch:
To do: handle converting links during unlinked->anonymous and anonymous->unlinked conversion

src/rard/research/models/fragment.py
```diff
@@ -212,6 +212,11 @@ def related_queryset(self):
     # the fragments that we are apposita for...
     fragments = models.ManyToManyField("Fragment", blank=True, related_name="apposita")
 
+    # the anonymous fragments that we are apposita for
+    anonymous_fragments = models.ManyToManyField(
+        "self", blank=True, symmetrical=False, related_name="anonymous_apposita"
```
acholyn, line 217: Why would it not be symmetrical? If you add AF2 as an apposita to AF1 is AF1 not also an apposita of AF2?

tcouch: Yeah, I asked Richard about that and he said it shouldn't be symmetrical. See #374

acholyn (approved):
Seems to work quite nicely. I wonder if we should just hide the Appositum Links section if there are none instead of displaying the heading and "None"

tcouch:
> Seems to work quite nicely. I wonder if we should just hide the Appositum Links section if there are none instead of displaying the heading and "None"

Yeah good point. We'd need to show it when in edit mode so users can access the buttons, but I'll hide it for the normal view.

tcouch:
@acholyn I've made that change to the template as suggested. Will merge now.

PR #442, opened 2024-03-14
Feature/consolidating edit menu

### LINKED ISSUE #434: consolidate action options into edit dropdown
The `Edit` dropdown could allow duplication, deletion, conversion and locking from one menu, with the former 3 being allowed so long as there isn't an existing lock on the item.

acholyn:
closes #434 and #447
---
The `Edit` menu has been rejigged to better display options and complete actions without having to require an editing lock. This helps bring more consistency and reduces the code volume.

- Deletion, conversion and duplication are available where relevant without needing a lock and are available when there is a lock
- locking editing is available from the same menu as a dropdown 

Below are screengrabs:

(unlinked) fragment without editing lock:
image

when editing:
image

testimonia, works and antiquarians when editing:
image

anonymous without editing lock:
image

### DIFF
--- a/src/rard/static/css/project.css
+++ b/src/rard/static/css/project.css
@@ -390,14 +390,6 @@ button.insert {
   background-color: #ffff00;
 }
 
-.lock-duration {
-  display: none;
-}
-
-.lock-options:hover .lock-duration {
-  display: block;
-}
-
 /* editor related */
 .ql-editor {
   background-color: white !important;
--- a/src/rard/static/css/project.css
+++ b/src/rard/static/css/project.css
@@ -390,6 +390,14 @@ button.insert {
   background-color: #ffff00;
 }
 
+.unlock-duration {
+  display: none;
+}
+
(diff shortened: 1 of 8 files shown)


tcouch (changes requested):
A few bugs here:
* I couldn't get Delete to work with or without an editing lock in place
* Warnings don't show up for make anonymous/unlinked or duplicate
* If no lock is in place, when you click make anonymous/unlinked you get a message saying you can't because there's no lock

You're not going to be happy with me but I think it might be better if we just have the consolidated 'Actions' menu in edit mode, and keep the original 'Edit' menu for when there isn't a lock.

acholyn:
I don't think we have any duplication warnings set up
Shall the menu be "edit" as before (when no lock) or still actions > edit

acholyn:
To do:
change it back so that the options are only available when there's a lock

acholyn:
need to update so testimonia have the option for duplication and update unlinked with option to convert to testimonium

Issue #452, opened 2024-05-10
Fix mentions interfering with TOF render


acholyn:
The quill mentions module seems to interfere with the rendering of the quill editor, though it isn't clear why.

Presently it's been disabled on production until we find a solution.

Potentially tangentally related - sometimes when adding text to a TOF that already has content, instead of appending to the main content, it gets appended via the `ql-cursor` element (eg. in Commentary of AF 53). Unsure if this is because the last elem is `<em>`

Issue #433, opened 2024-02-27
improve mentions response time


acholyn:
Mentions are quite slow, particularly on local. It may be worth adjusting the request such that it should only be submitted after `:` and x milliseconds later

Might be worth leveraging our ElasticSearch through django haystack and use this for mentions

acholyn:
related to #454 
Might be best to see if researchers have issues with current system before overhauling

PR #453, opened 2024-05-13
Switch to CKEditor

### LINKED ISSUE #423: Issues with pasting into text editor triggering footnote section via carriage returns

tcouch:
closes #423 and #452 
related to #433 , may have resolved it since it caches now
---
Trying CKEditor as a fix for Quill's broken mention plugin (plus the weird pasting behaviour too hopefully)


src/rard/research/forms.py
```diff
@@ -67,6 +67,9 @@ def __init__(self, *args, **kwargs):
             self.fields[
                 "introduction_text"
             ].initial = self.instance.introduction.content
+            self.fields["introduction_text"].widget.attrs[
+                "class"
+            ] = "enableMentions enableCKEditor"
```
tcouch: At the moment enableFootnotes is only applied to the Commentaries, should they be enabled on Introductions (Antiquarian, Work, Book...) as well? I can't remember.

tcouch:
It just occurred to me that the html for existing footnotes won't be compatible with the new footnotes style. We'll need to write a migration to convert them.

tcouch:
The introduction on WorkForm and BookForm needs to have the editor enabled.

tcouch:
Public commentary inline edit form needs to have its editor enabled

tcouch:
Commentary apparatus criticus mentions aren't working ~if there are two original texts~

src/rard/research/views/__init__.py
```diff
@@ -89,7 +89,7 @@
     fetch_fragments,
 )
 from .history import HistoryListView
-from .home import HomeView, render_editor_modal_template
```
tcouch, line 92: This doesn't ring a bell at all. Any idea what it was?

src/rard/research/views/fragment.py
tcouch, line 1: The duplicate_fragment view function is massive. Most of the logic should probably be split out into utils and possible divided into a few smaller functions.

src/rard/static/js/app-crit-editor.js
```diff
@@ -0,0 +1,271 @@
...
+      }
+    });
+  });
+}
+
+function updateOriginalTextReferences() {
+  var originalTextEditor = CKEDITOR.instances["id_content"];
+  var content = originalTextEditor.getData();
+  var mentions = content.match(/data-id="\d+"/g);
+  var refSpans = content.match(/<span><span>#<\/span>\d+<\/span>/g);
```
tcouch: I get refSpans = null error when it gets to this function. Maybe the regex isn't quite right?

On the one I'm looking at, there are spaces around the number, so this might cover either case:
`/<span><span>#<\/span>\s*\d+\s*<\/span>/g`

src/rard/static/js/project.js
```diff
@@ -883,244 +589,3 @@ function runMoveAction(post_data, post_url) {
...
-    let headers = {};
-    headers["X-CSRFToken"] = csrf;
-
-    // any text within the original text editor on the same
-    // page might have become stale due to a change in the apparatus criticus
-    // so we send it to the server to re-index any app crit links within it.
-    // nb this doesn't save anything, just refreshes the text in the editor
-    // so that it appears correct to the user. It would be all handled correctly
-    // on saving anyway, but this is a cosmetic update so the user doesn't become confused!
-    $.ajax({
```
tcouch, line 941: I guess this approach is an alternative to what you were doing with the regex in app-crit-editor.js, but I think your way is better. If I understand correctly, this would update the app crit indexes at the expense of throwing away without warning any other changes the user might have made to the original text.

src/rard/research/forms.py
acholyn: seems issue #392 says it should be, I'll update that

acholyn:
> The introduction on WorkForm and BookForm needs to have the editor enabled.

do they not use the IntroductionFormBase? Do I also need to enable it on the AntiquarianCreateForm as well?

acholyn:
> Public commentary inline edit form needs to have its editor enabled

https://github.com/UCL/frrant/blob/7c753c73f94358610d963b424ca50ed7368b2b4e/src/rard/research/forms.py#L702
is this in the wrong place?

src/rard/research/views/__init__.py
acholyn: that was because I had to do a modal for the editor for like footnotes

tcouch:
> > The introduction on WorkForm and BookForm needs to have the editor enabled.
> 
> do they not use the IntroductionFormBase? Do I also need to enable it on the AntiquarianCreateForm as well?

Apparently not - it's the form you get when you create a new work or book as opposed to editing them once they're created. And yes looks like AntiquarianCreateForm also needs it.

tcouch:
> > Public commentary inline edit form needs to have its editor enabled
> 
> https://github.com/UCL/frrant/blob/7c753c73f94358610d963b424ca50ed7368b2b4e/src/rard/research/forms.py#L702
> 
> is this in the wrong place?

Pushed a fix: I think the problem was it was inside an if block that didn't always get executed. Copied the approach used for CommentaryFormBase instead.

src/rard/research/views/fragment.py
acholyn: have opened #455 to address this in a later PR as another PR currently updates the logic

tcouch:
The create forms for Works and Antiquarians (`/work/create/` and `antiquarian/create/`) need to have their editors enabled.

PR #456, opened 2024-06-03
Feature/duplicating testimonia

### LINKED ISSUE #450: Duplicating testimonia
Expansion of #393 to include testimonia
- [x] allow conversion of unlinked to testimonium
- [x] expand duplicates list for duplicate testa
- [x] enable button 
- [x] update tests
  - [x] duplication
  - [x] conversion

acholyn:
closes #450 
---


src/rard/research/migrations/0071_add_duplicates_field_squashed.py
```diff
@@ -1,10 +1,14 @@
...
 
 from django.db import migrations, models
 
 
 class Migration(migrations.Migration):
 
+    replaces = [
+        ("research", "0071_add_duplicates_field"),
+    ]
+
```
tcouch, line 11: I don't think this will work. If I go to the development branch and run migrations I'm on 0071, then when I switch to this branch and run migrations nothing happens because it thinks its already run 0071. The only way to get it to work is to revert to 0070 while on the development branch, then switch to feature/duplicate-testimonia and run the migration. That won't work on production though where we've already merged the changes into the production branch.

tcouch: I don't think this will work. If I go to the development branch and run migrations I'm on 0071, then when I switch to this branch and run migrations nothing happens because it thinks its already run 0071. The only way to get it to work is to revert to 0070 while on the development branch, then switch to feature/duplicate-testimonia and run the migration. That won't work on production though where we've already merged the changes into the production branch.

src/rard/templates/research/testimonium_detail.html
tcouch, line 1: This template seems to be missing the duplicates list at the bottom

src/rard/utils/convertors.py
```diff
@@ -11,28 +12,32 @@ class FragmentIsNotConvertible(Exception):
...
-        # for each duplicate create a new relationship to the frag
-        for d in source.duplicates_list:
-            d.duplicate_frags.add(destination)
-            # if the relationship was defined from the duplicate in the list, remove the one being converted
-            if source in d.duplicate_afs.all():
-                d.duplicate_afs.remove(source)
-            else:
-                # otherwise delete it from the anonfrag being converted
-                source.duplicate_afs.remove(d)
+def transfer_duplicates(source, destination):
```
tcouch, line 15: Updated in recent commit to ensure destination is added as duplicate of the source's duplicate as well as the reverse relationship being created.

src/rard/research/models/base.py
```diff
@@ -571,18 +571,39 @@ def mentioned_in_list(self):
...
+    # if converted to a testimonium, we want to transfer duplicates
+    duplicate_ts = models.ManyToManyField(
+        "Testimonium",
+        blank=True,
+        related_name="%(class)s_duplicate_ts",
+    )
+
     @property
     def duplicates_list(self):
         duplicates = list(self.duplicate_frags.all())
```
tcouch, line 583: I've simplified this in the latest commit: as long as we make sure duplicate relationships are always bi-directional, we shouldn't have to look at the reverse relationships.

tcouch (changes requested):
Looks good except for missing duplicates section on Testimonium detail page. Also delete button seems not to work, but I guess that's a different PR.

acholyn:
> Looks good except for missing duplicates section on Testimonium detail page. Also delete button seems not to work, but I guess that's a different PR.

think it's in #442 (re button)
have also opened a new one specifically for this since 442 automatically merged with this PR

tcouch (approved):

PR #462, opened 2024-06-13
Amending translation public view

### LINKED ISSUE #37: Non-chosen translations should be hidden from public view

acholyn:
closes #37
---
Only approved translations shown to the general public (without object lock), this is without the green border or "approved" note (irrelevant if only one seen)
This can probably be merged later on


tcouch (approved):
Looks good. Not noticed that issue before though. Are the research team aware that you've got a solution for this? Check with them that they're happy with the solution before applying to prod since it'll make translations disappear from the public view.

PR #464, opened 2024-06-18
fixing delete not working in new menu format

### LINKED ISSUE #434: consolidate action options into edit dropdown

acholyn:
closes #434 
---
#442 automatically merged when #456 did so have done this fix separately

The delete form wasn't rendering  because of the nesting within the bigger form so I've split the menu into 2 versions, one for when it's locked, and one for when it's unlocked, with the relevant forms

### DIFF
--- a/src/rard/templates/research/partials/render_locked_info.html
+++ b/src/rard/templates/research/partials/render_locked_info.html
@@ -9,8 +9,6 @@
 
         </div>
         <div>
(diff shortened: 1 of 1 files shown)


tcouch (approved):
Looks good to me.

PR #472, opened 2024-08-06
Feature/new concordances

### LINKED ISSUE #446: New concordance structure
Concordances will have a new structure, which will still essentially be a string but will be composed of different components in dropdown menus in order to construct it. There should also be some charfields to allow for outlier reference patterns. See the [document](https://github.com/UCL/frrant/files/14704144/Concordance.wireframe.docx) for a fleshed out wireframe regarding creation and searching.

The detail view should be like the search view, the items on the page are only displayed when a search has been done and the entries can be edited in place.

One aspect that will need further details is displaying the search results where there's no concordance available

acholyn:
closes #446 
---
- new `ConcordanceModel`, `Edition` and `PartIdentifier` models
- `PartIdentifiers` belong to `Editions` and a `ConcordanceModel` has an identifier (`PartIdentifier`) and edition (through `PartIdentifier`)
- a `ConcordanceModel` has an associated `OriginalText`
- old concordances are still displayed and can be deleted but not edited or updated, they are only there for reference so researchers know what new ones need to be made
- The `ConcordanceListView` is different now and requires a search before results are returned, no pagination (specifically requested)


src/rard/templates/research/partials/section/original_text_details.html
```diff
@@ -52,12 +52,31 @@
...
-            {% if original_text.concordances.count == 0 %}
+            <div>
+            {% if original_text.concordances.count != 0 %}
+                <em>{% trans 'old concordances:' %}</em>
+                {% for c in original_text.concordances.all%}
+                <div class='d-flex justify-content-between mb-3'>
+                <form novalidate class='form-inline' action='{% url "concordance:delete_old" c.pk %}' method='POST'>
+                    {% csrf_token %}
+                    <small>{{c}}</small>
+                        {% if perms.research.delete_concordance %}
```
tcouch: Delete button should only be shown is there's a lock on the fragment

src/rard/templates/research/partials/section/original_text_details.html
```diff
@@ -52,12 +52,31 @@
...
+                <em>{% trans 'new concordances:' %}</em>
+                {% if original_text.concordance_models.count == 0 %}
                 None
+                {% endif %}
             {% endif %}
-            {% for concordance in original_text.concordances.all %}
+            {% for concordance in original_text.concordance_models.all %}
                 <div class='d-flex justify-content-between mb-3'>
-                <div>{{ concordance }}</div>
+                <div>{{ concordance }} </div>
```
tcouch, line 89: Where the Concordance doesn't have a part identifier I'd leave out the [none] part
image

src/rard/research/models/concordance.py
```diff
@@ -0,0 +1,94 @@
...
+from rard.research.models.mixins import HistoryModelMixin
+from rard.utils.basemodel import BaseModel
+
+
+class PartIdentifier(HistoryModelMixin, BaseModel):
+    class Meta:
+        ordering = ["edition__name", "value"]
+
+    def __str__(self):
+        return f"{self.edition.name}: {self.value}"
```
tcouch: Possibly edit this to just return edition name if value is none?

src/rard/templates/research/partials/section/original_text_details.html
```diff
@@ -52,12 +52,31 @@
     <dl class="row mb-0">
         <dt class="col-sm-3">{% trans 'Concordances' %}</dt>
         <dd class="col-sm-7">
-            {% if original_text.concordances.count == 0 %}
+            <div>
+            {% if original_text.concordances.count != 0 %}
+                <em>{% trans 'old concordances:' %}</em>
```
tcouch: Layout suggestion: put old concordances in a separate `<dl>` below the new concordances?

src/rard/templates/research/concordancemodel_form.html
tcouch, line 1: The layout seems a bit strange here - is it just my browser?
image

In particular the indentation of the form with respect to the content above, and the larger sans serif font for "Create or select an Edition" seems a bit off compared to the "Create Concordance" title at the top. And the second form row is squished.

tcouch: Could also do with a Cancel button.

src/rard/research/tests/forms/test_concordance.py
tcouch, line 1: Delete this if there's nothing to go in it.

src/rard/templates/research/concordancemodel_form.html
tcouch, line 1: If I create a new edition, then click through to create a new part identifier I get the following error:
ValueError: Field 'id' expected a number but got ''.

image
image

tcouch (changes requested):
I haven't gone through all of it yet, but found a few issues.

src/rard/research/models/concordance.py
acholyn: we wouldn't want that in all instances since seeing the format is necessary in some instances. I think I've figured out how to do both though

src/rard/templates/research/concordancemodel_form.html
acholyn: the indentation was intentional so that things are centred, should I remove that?. Second row shouldn't look like that 🤔 
image

acholyn: hm, it seems the `new_edition` pk isn't getting passed over properly

tcouch: How about:

```
Select an Edition:

<dropdown>

Or, create a new edition:

<name> <full name>
```

src/rard/templates/base.html
```diff
@@ -123,6 +123,8 @@
 
       <script src="{%static 'js/ckeditor/ckeditor.js'%}"></script>
 
+      <script src="https://cdnjs.cloudflare.com/ajax/libs/fuse.js/6.4.6/fuse.min.js"></script>
```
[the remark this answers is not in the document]
tcouch: @acholyn if we want to address this we can use this site to generate an integrity hash: https://www.srihash.org/

I also noticed the fuse website uses this url format to specify the version: 
`<script src="https://cdn.jsdelivr.net/npm/fuse.js@7.0.0"></script>`

...though I'm not sure if it matters.

src/rard/templates/research/partials/section/original_text_details.html
tcouch: Is still think this would look better as a separate table row. Do you disagree?

src/rard/templates/research/concordancemodel_form.html
tcouch: I'm still seeing the same squished layout

tcouch: ...but only when I use my monitor's full width. It actually starts to look ok as I make the window narrower.

tcouch:
I've found a new issue when creating new concordances: if I try to create a concordance where the edition doesn't have a part identifier I get an error: `rard.research.models.concordance.PartIdentifier.DoesNotExist: PartIdentifier matching query does not exist`

src/rard/templates/research/concordancemodel_form.html
acholyn: 🤔
what's the full width of your monitor? I may need to add a larger breakpoint

tcouch: window.innerWidth = 1920

acholyn: have pushed some tweaks that hopefully work but do let me know if not. Relevant to the creation form and the search page

src/rard/research/models/concordance.py
```diff
@@ -0,0 +1,102 @@
...
+            return f"{self.name} ({self.description})"
+
+    name = models.CharField(max_length=128, blank=False)
+
+    description = models.CharField(
+        max_length=200, blank=True, null=True
+    )  # need to add this to master bib
+
+
+class ConcordanceModel(HistoryModelMixin, BaseModel):
```
tcouch, line 60: Do we need to define ordering for this model on a Meta class? I presume that's what the concordance_order field is for?

tcouch, line 60: I can't see a display_order field here, but it's on the form. What happens to that?

src/rard/templates/research/concordancemodel_form.html
tcouch, line 1: I find this form a bit confusing where it presents both options at once to create a new edition or select an existing one. A better approach might be to present the select dropdown by default with a button to allow users to create a new one instead. On clicking the button, the alternative form would replace the dropdown. You can use the same approach for both editions and part identifiers.

tcouch: Also, would it make sense for an empty response in the format field to default to [none]? It'd save people the trouble of having to write "[none]" and potentially getting it wrong.

src/rard/research/models/concordance.py
acholyn: at present not 100% sure if the concordance_order should override the ordering I was told about for the search results (see get_ordered_queryset()) or if it should be applied after

src/rard/templates/research/concordancemodel_form.html
tcouch: Here's a mock up of what I mean (I should probably have used Figma)
image

src/rard/research/models/concordance.py
acholyn: it's attached to the `PartIdentifier` model since that's what it applies to

src/rard/templates/research/concordancemodel_form.html
acholyn: shall we put it on preprod and see if the researchers would prefer that? My thinking in having them both available initially was to reduce the number of clicks needed to create a new one - they can check the select dropdown to see if the edition is already there and if not they can just start making immediately

acholyn: > Also, would it make sense for an empty response in the format field to default to [none]? It'd save people the trouble of having to write "[none]" and potentially getting it wrong.

Potentially, yes -that makes sense to me

tcouch: And for the part identifier:
image

tcouch: I think this makes it clearer that display order is a property of the part identifier rather than the concordance (or have I misunderstood?)

tcouch: Where the edition is new and no part identifiers exist yet, show the new part identifier form section, otherwise default to select.

tcouch, line 1: There are a few issues here:
* If I click "create a new edition", the "Edition full name" and "format of parts" inputs are disabled
* New part identifier is disabled when edition format is [none], but display order isn't
* Concordance detail inputs are indented relative to part identifier inputs
* Concordance form section shows this field is required errors even though it's the first time it's shown
* Could do with a bit more space and a heading between the part identifier and concordance detail sections

acholyn: Any ideas for how to stop the required errors without disabling them entirely?
I think it happens because technically it's handling a post to render it

src/rard/research/views/concordance.py
```diff
@@ -82,75 +212,144 @@ class ConcordanceCreateView(
...
+    def get(self, request, *args, **kwargs):
+        edition_form = EditionForm()
+        concordance_form = None
+        return render(
+            request,
+            self.template_name,
+            context=self.get_context_data(edition_form, concordance_form),
+        )
+
+    def post(self, request, *args, **kwargs):
```
tcouch, line 283: I think you're trying to put too much functionality into one form here, which is why we end up with things like the error messages appearing when the second stage gets shown. Why not use separate edition create and concordance create forms with two separate views for each stage? For example:

* Edition view (stage 1): User selects an edition, or
* User selects create new edition and completes the relevant form
* User clicks next and either:
  * The concordance create view (stage 2) is called with a get request passing the selected edition as a query parameter, or
  * The edition create form is posted to the edition view, which creates the new edition and then the success url redirects to the concordance create view passing the newly created edition as a parameter
* The concordance create view returns a concordance form instance, passing the edition id in order to figure out the part identifier options
* On submit, the concordance view still handles both creating part_identifiers and concordances, but maybe split this functionality into separate functions to make the high level logic of the post function a bit clearer

src/rard/templates/research/concordancemodel_form.html
tcouch: Sorry, just saw your comment. See my later comment on the concordance view for a suggestion.

tcouch:
To recap our discussion: PartIdentifier display_order should be set equal to "value" if nothing is entered when the form is submitted.

src/rard/research/views/concordance.py
```diff
@@ -82,75 +253,165 @@ class ConcordanceCreateView(
...
+                    edition=Edition.objects.get(pk=edition),
+                    value=concordance_form.cleaned_data["new_identifier"],
+                    display_order=concordance_form.cleaned_data["display_order"],
+                )
+                new_identifier_instance.save()
+                identifier = new_identifier_instance.pk
+            elif identifier is None:
+                identifier = get_part_format(edition).pk
+
+            if display_order is None:
```
tcouch: Did you check this works? I think it isn't working for a couple of reasons: 
* line 290 where you get display_order from the POST data actually returns an empty string rather than None, so this if block never gets called - empty strings evaluate to false though, so you could just use `if not display_order:`
* the other thing is that you're not actually doing anything with display_order here after updating the value.

I'd usually suggest using a Form's clean method to handle this sort of situation (see the Django docs on [form validation](https://docs.djangoproject.com/en/5.1/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other)). When you call is_valid() on the concordance form above, it should run the clean method on the form and update the value.

tcouch:
FYI my latest commit edited migration 0073 to address the fact I'd removed the meta options from PartIdentifier. I don't think it actually has any effect on the database, but if you do run into any issues or just want to be safe, you'll need to roll back to 0072 and apply 0073 again.

tcouch (approved):

PR #474, opened 2024-08-21
closes #471

### LINKED ISSUE #471: 500 error caused by app crit search within original text create form
**Describe the bug**
The text editor on the original text create form (used in fragment, anonymous fragment, and testimonium create views) has the mentions plugin enabled for apparatus criticus searching. This sends a get request to `/search/ajax/apparatus-criticus/` whenever the hash key '#' is pressed. Unfortunately, this request needs to include an object_id as one of the parameters, but object_id=null because this is a yet to be created object. As a result we get a 500 error.

It's not a massive problem in terms of user experience as you'd only actually see the ajax request response if you are looking at the network pane, but it does mean we get emails when it happens in production.

Recommendation: App crit search should probably be disabled on this form.

acholyn:
Should fix issue as the OT shouldn't have any ap crit items on creation

### DIFF
--- a/src/rard/research/forms.py
+++ b/src/rard/research/forms.py
@@ -572,9 +572,9 @@ class OriginalTextDetailsForm(forms.ModelForm):
         if original_text:
             self.fields["content"].widget.attrs["data-object"] = original_text.pk
             # Only enable apparatus criticus editing if object exists
-            self.fields["content"].widget.attrs[
-                "class"
-            ] = "enableApparatusCriticus enableCKEditor"
+            self.fields["content"].widget.attrs["class"] = "enableCKEditor"
+            if original_text.apparatus_criticus_items.count() > 0:
+                self.fields["content"].widget.attrs["class"] = "enableApparatusCriticus"
 
     def clean_reference_order(self):
         # Reference order needs to be stored as a string with leading 0s such
--- a/src/rard/research/forms.py
+++ b/src/rard/research/forms.py
@@ -621,7 +621,8 @@ class OriginalTextForm(OriginalTextAuthorForm):
         original_text = kwargs.get("instance", None)
         if original_text and original_text.pk:
             original_text.update_content_mentions()
-
+            if original_text.apparatus_criticus_items.count() > 0:
+                self.fields["content"].widget.attrs["class"] = "enableApparatusCriticus"
         if original_text and original_text.reference_order:
             original_text.reference_order = (
                 original_text.remove_reference_order_padding()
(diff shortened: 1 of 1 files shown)


src/rard/research/forms.py
```diff
@@ -572,9 +572,9 @@ def __init__(self, *args, **kwargs):
         if original_text:
             self.fields["content"].widget.attrs["data-object"] = original_text.pk
             # Only enable apparatus criticus editing if object exists
-            self.fields["content"].widget.attrs[
-                "class"
-            ] = "enableApparatusCriticus enableCKEditor"
+            self.fields["content"].widget.attrs["class"] = "enableCKEditor"
+            if original_text.apparatus_criticus_items.count() > 0:
+                self.fields["content"].widget.attrs["class"] = "enableApparatusCriticus"
```
tcouch: I haven't tested this in practice, but won't this set the class to "enableApparatusCriticus" rather than "enableCKEditor enableApparatusCriticus" if there any apparatus criticus items? If so, I'd expect the editor not to get initialised. Likewise for OriginalTextForm below.

tcouch (changes requested):

src/rard/research/forms.py
acholyn: ah, can we use += or will we just have to repeat the ckeditor string?

src/rard/research/forms.py
```diff
@@ -574,7 +574,9 @@ def __init__(self, *args, **kwargs):
             # Only enable apparatus criticus editing if object exists
             self.fields["content"].widget.attrs["class"] = "enableCKEditor"
             if original_text.apparatus_criticus_items.count() > 0:
-                self.fields["content"].widget.attrs["class"] = "enableApparatusCriticus"
+                self.fields["content"].widget.attrs[
+                    "class"
+                ] += "enableApparatusCriticus"
```
tcouch: You'll need a space
```suggestion
                ] += " enableApparatusCriticus"
```

src/rard/research/forms.py
```diff
@@ -622,7 +624,9 @@ def __init__(self, *args, **kwargs):
         if original_text and original_text.pk:
             original_text.update_content_mentions()
             if original_text.apparatus_criticus_items.count() > 0:
-                self.fields["content"].widget.attrs["class"] = "enableApparatusCriticus"
+                self.fields["content"].widget.attrs[
+                    "class"
+                ] += "enableApparatusCriticus"
```
tcouch:
```suggestion
                ] += " enableApparatusCriticus"
```

src/rard/research/forms.py
```diff
@@ -622,7 +624,9 @@ def __init__(self, *args, **kwargs):
         if original_text and original_text.pk:
             original_text.update_content_mentions()
             if original_text.apparatus_criticus_items.count() > 0:
```
tcouch: Is there a reason this needs to come before the call to super().__init__ ? Just that it probably makes sense to follow the same order as above where `ennableApparatusCriticus` is added to the class list after `enableCKEditor`. Also I don't know if += would work here if widget.attrs["class"] potentially doesn't exist as a key yet. Have you tested it?

tcouch: Ah yeah, the failing pytest check seems to confirm this.

tcouch (approved):

PR #475, opened 2024-09-04
Feature/new concordances tc fixes

tcouch:
First of all fixes a bug caused by existing edition's PI templates not getting picked up. The solution uses a custom manager to include is_template as an annotation, which allows us to filter and order by is_template directly. It did mean I had to use a regex in the __str__ function though because you can't have a property and an annotation with the same name. I think the simplest solution would have been for is_template to have been a boolean field on the PartIdentifier model, but that'd involve another migration.

I've also made a few tweaks to styling and form layout - feel free to revert these or cherrypick the good bits.

### DIFF
--- a/src/rard/static/css/project.css
+++ b/src/rard/static/css/project.css
@@ -527,7 +527,7 @@ section.footnotes {
 }
 
 .form-section {
-  margin-bottom: 40px;
+  margin-bottom: 3rem;
 }
 
 /* concordance search */
--- a/src/rard/static/css/project.css
+++ b/src/rard/static/css/project.css
@@ -490,7 +490,7 @@ section.footnotes {
 }
 .concordance-form__area {
   min-height: 30vh;
-  margin-inline: 2vw;
+  /* margin-inline: 2vw; */
 
   [disable-selection="true"] {
     color: gray;
@@ -524,6 +524,10 @@ section.footnotes {
     bottom: 2.5rem;
     left: 8rem;
   }
+}
+
+.form-section {
+  margin-bottom: 40px;
 }
 
 /* concordance search */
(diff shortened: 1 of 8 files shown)


src/rard/static/css/project.css
```diff
@@ -526,6 +526,10 @@ section.footnotes {
   }
 }
 
+.form-section {
+  margin-bottom: 40px;
```
acholyn: can we do this as vh or rem por favor 🙏

src/rard/templates/research/partials/concordance_edition_section.html
```diff
@@ -5,7 +5,7 @@
 
     <div class="d-flex flex-column justify-content-between edition-select">
     {% bootstrap_field edition_form.edition %}
-    <h2 class="alphabetum btn-link new-edition-btn">Or create a new Edition >></h2>
```
acholyn, line 8: do we not need it to be in alphabetum anymore?

acholyn (approved):

src/rard/static/css/project.css
tcouch: Fixed - used rem

src/rard/templates/research/partials/concordance_edition_section.html
tcouch: I think the convention across the rest of the site is for titles and form labels to use alphabetum, but additional instructions use the sans-serif font. In any case I found the alphabetum class was overriding the small class because it sets font size to be large. Happy to be overruled though.

acholyn: I'm happy for it to be sans-serif

PR #477, opened 2024-09-25
Add proximity search

### LINKED ISSUE #299: Proximity search with tilde character
From #297 "Another useful feature would be the ability to run proximity searches (i.e. find occurrences of two or more words within a certain number of words / characters of one another; I think PHI Latin do this using the ~ symbol, e.g. sed~et finds instances of these words within a set number of characters [and in that order]"

tcouch:
closes #299 

As described in the issue, this introduces the ability to carry out a proximity search where a search string like:

`Tarquinium ~5 ferens`

Would match: 

`Tarquinium Superbum regem adiit nouem libros ferens`

In the example above `~5` represents exactly five intervening words. You can also specify upper and lower bounds for the number of intervening words:

* `Tarquinium ~:5 libros` matches because `:5` means up to five words
* `Tarquinium ~3:5 libros` matches because `3:5` means between three and five words, but `Tarquinium ~3:5 adiit` would not match because there are only 2 intervening words

### DIFF
--- a/src/rard/research/tests/views/test_search.py
+++ b/src/rard/research/tests/views/test_search.py
@@ -444,6 +444,46 @@ class TestSearchView(TestCase):
         self.assertEqual(do_search(view.fragment_search, "again"), [f3])
         self.assertEqual(do_search(view.fragment_search, "again?"), [])
 
+    def test_proximity_search(self):
+        # Run a particular search and return a list of results
+        def do_search(search_function, keywords):
+            return list(search_function(SearchView.Term(keywords)))
+
+        cw = CitingWork.objects.create(title="citing_work")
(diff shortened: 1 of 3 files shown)


acholyn:
Seems to work for whole words nicely. The issue mentions also working with incomplete words eg `Marc ~1 varro` should still return Marcus Varro but it doesn't return anything currently

acholyn:
Would be good to update the helptext to reflect that the ambiguity is possible but you need to add the wildcard character (I didn't realise this)

tcouch:
@acholyn I've updated the help text as well as fixing another issue which was stopping search snippets from working. Do you want to have another look before I merge?

acholyn (approved):

PR #478, opened 2024-09-26
adding introduction to citing author and citing work

### LINKED ISSUE #383: Adding text boxes to citing authors and citing author works
@tcouch Hi Tom, would it be possible to add a text box to each citing author or (preferably) each citing author's work? If so, could bibliography be captured as for antiquarians?

acholyn:
closes #383


tcouch (changes requested):
When adding or updating an introduction on a citing_work or citing_author, if it includes a mention of a bibliography item, you get the following error:

```
File "/app/rard/utils/basemodel.py", line 319, in link_bibliography_mentions
     for ant in antiquarians:
UnboundLocalError: local variable 'antiquarians' referenced before assignment
```

We can't really assign an antiquarian to a citing author or citing work, but we would like to link bibliography items in the same way. We'll have to expand the `link_bibliography_mentions` method to handle this.

src/rard/utils/basemodel.py
```diff
@@ -303,6 +303,12 @@ def link_bibliography_mentions(self):
                 if self.book:
                     antiquarians = {ant for ant in self.book.work.antiquarian_set.all()}
 
+                if self.citing_work:
+                    antiquarians = {self.citing_work.author}
```
tcouch: This works, but I think it's cheating. Since this variable no longer necessarily refers exclusively to Antiquarian model instances, I think it needs renaming to something like `bibliography_owners`.

src/rard/templates/research/citingauthor_detail.html
tcouch, line 1: If you look at the Antiquarian detail page, there's a htmx trigger on the bibliography item list which refreshes it when the introduction is updated. It just means you get to see the updated bibliography without having to refresh the page.

src/rard/research/views/bibliography.py
```diff
@@ -173,6 +174,8 @@ def get_queryset(self) -> QuerySet[Any]:
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+        if self.ant_pk:
+            context["antiquarian"] = Antiquarian.objects.get(id=self.ant_pk)
```
tcouch: Repeated lines here. Was this supposed to do something different?

src/rard/research/views/bibliography.py
```diff
@@ -163,6 +163,7 @@ class BibliographySectionView(LoginRequiredMixin, PermissionRequiredMixin, ListV
     context_object_name = "bibliography_items"
     permission_required = ("research.view_bibliographyitem",)
 
+    # todo: update this to dynamically work with authors and ants
```
tcouch: This needs doing doesn't it? You could use this approach to pass kwargs depending on the URL used to call the view:
https://docs.djangoproject.com/en/5.1/topics/http/urls/#views-extra-options

src/rard/research/urls.py
```diff
@@ -649,6 +649,26 @@
...
+                    path(
+                        "<pk>/update/introduction/",
+                        views.CitingAuthorUpdateIntroductionView.as_view(),
+                        name="update_introduction",
+                    ),
+                    path(
+                        "<pk>/bibliography/",
+                        views.BibliographySectionView.as_view(),
+                        name="bibliography",
+                    ),
```
tcouch: If using the [passing extra options](https://docs.djangoproject.com/en/5.1/topics/http/urls/#views-extra-options) approach, I think this could work:
```suggestion
                    path(
                        "<pk>/bibliography/",
                        views.BibliographySectionView.as_view(),
                        {"related_model":"citing_author"},
                        name="bibliography",
                    ),
```

src/rard/research/views/bibliography.py
tcouch: Currently this view uses `research/partials/antiquarian_bibliography_list.html` which includes a htmx process for refreshing the content that points to the antiquarian:bibliography url. You'll need to either switch to a different template depending on the related model in the view, or pass something in the context data that can be used to switch the url in the template.

tcouch (changes requested):

src/rard/research/views/bibliography.py
acholyn: I've bypassed the need for the `antiquarian_bibliography_list` template with citing authors by including the equivalent in the detail template. My thinking was that htmx doesn't need a whole new template to update the content.
I think I've gotten confused somewhere along the way and am trying to blend two different strategies to this solution

src/rard/research/views/bibliography.py
acholyn: May be a holdover from when I was going to try something and then either it needs to be updated or is no longer relevant

acholyn:
I've also just noticed that we don't have a shortcode for citing authors in mentions, so how do we mention them? Or is this just another thing to do for this PR

tcouch:
> I've also just noticed that we don't have a shortcode for citing authors in mentions, so how do we mention them? Or is this just another thing to do for this PR

I guess the researchers have just never needed to mention citing authors. Is there a reason we would want to add it?

acholyn:
>Is there a reason we would want to add it?

Perhaps not, I suppose the researchers can always request this in future

acholyn:
Possible qol: when clicking the refresh, it should do a loader since typically on a CA it will already be updated and won't actually change

src/rard/research/views/bibliography.py
tcouch: The problem is that updating the intro triggers the bibliography to refresh via this view, and the html content that gets swapped in has two problems:
1. No bibliography items are returned because the view can't handle citing authors at the moment
2. The template has `hx-get="{% url 'antiquarian:bibliography' pk=antiquarian.pk %}"` in it, but we need that to point to the citing_author:bibliography url if it's being rendered for the citing author detail page.

tcouch: To replicate the issue I'm talking about:
1. Edit a citing author's introduction including a mention of a bibliography item
2. Save the changes
3. You should see the bibliography section at the bottom automatically update, but without any content
4. Note the url in the #bib-list element at the bottom of the page

acholyn: Sorry, changes hadn't been pushed! Think these have been addressed in the latest one; I've used a `get_templates` function so it points to the right one and put the citing author section in a new template

tcouch (approved):

PR #481, opened 2025-01-08
Using htmx indicator to control loader display

tcouch:
There's a htmx way to control the loader display which is probably better to use since it's tied to the request (see [hx-indicator](https://htmx.org/attributes/hx-indicator/)).

I couldn't find your loader so I used a new one, but feel free to replace with something you think looks better. htmx uses the id to control display so there's no reason to have it inline with the button if you think overlaying it looks better.

### DIFF
--- a/src/rard/static/css/project.css
+++ b/src/rard/static/css/project.css
@@ -538,54 +538,39 @@ section.footnotes {
     max-width: 50%;
   }
 }
-.loader-wrapper {
-  display: flex;
-  justify-content: center;
-}
-#loader {
-  position: absolute;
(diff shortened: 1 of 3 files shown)


src/rard/static/css/project.css
acholyn, line 1: Is there a way to accept changes excluding a file so we can just keep the original loader?

acholyn:
have applied and tweaked in https://github.com/UCL/frrant/pull/478/commits/74cd19837e77cac99947696f5aa5b4e0df5d3f07

PR #482, opened 2025-01-20
Feature/381-accordion-display

### LINKED ISSUE #381: Display work links in accordion
On an antiquarian page, to encourage a cleaner look, we can display the works as accordions and only display the fragments for that work when the accordion is expanded. There should be a `expand all` button for ease of expanding all for antiquarians who do not have a lot of content.

acholyn:
On an Anqituarian's detail page, the works under `Ordered Material` should now have the work titles and a button with "expand" and a downward chevron. Clicking this button will expand the contents of the work.
There is also a button to expand/collapse all, which works by toggling. This is done with bootstrap.

image

closes #381

### DIFF
--- a/src/rard/templates/research/partials/ordered_work_area.html
+++ b/src/rard/templates/research/partials/ordered_work_area.html
@@ -34,11 +34,11 @@
                         {% if work.unknown == True and not flinks and not tlinks and not alinks %}
                         {% elif work.unknown == True %}
                         <b class='my-3 mr-1'><a href='{% url "work:detail" work.pk %}' >{{ work.name }}</a></b>
-                        <button type='button' class="btn btn-text text-primary pl-2" data-toggle="collapse" data-target="#list_{{work.pk}}" role="button" aria-expanded="false" aria-controls="list_{{work.pk}}"><i class='fas fa-caret-down'></i></button>
+                        <button type='button' class="btn btn-text text-primary pl-2 collapse-button" data-toggle="collapse" data-target="#list_{{work.pk}}" role="button" aria-expanded="false" aria-controls="list_{{work.pk}}"><i class='fas fa-chevron-down'><small class='ml-1 text-muted'>{% trans "Expand" %}</small></i></button>
 
                         {% else %}
                         <b class='my-3 mr-1'><a href='{% url "work:detail" work.pk %}'>{{ work.name }}</a></b>
-                        <button type='button' class="btn btn-text text-primary" data-toggle="collapse" data-target="#list_{{work.pk}}" role="button" aria-expanded="false" aria-controls="list_{{work.pk}}"><i class='fas fa-caret-down'></i></button>
+                        <button type='button' class="btn btn-text text-primary collapse-button" data-toggle="collapse" data-target="#list_{{work.pk}}" role="button" aria-expanded="false" aria-controls="list_{{work.pk}}"><i class='fas fa-chevron-down'><small class='ml-1 text-muted'>{% trans "Expand" %}</small></i></button>
 
                         {% if can_edit and has_object_lock %}
                         <button type='button' class='drag-handle btn btn-light btn-sm' style='cursor: move;' data-pos='{{ forloop.counter0 }}'><i class='fas fa-bars'></i></button>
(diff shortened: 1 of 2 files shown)


src/rard/templates/research/partials/ordered_work_area.html
```diff
@@ -105,3 +109,43 @@
 </div>
 
 {% endwith %}
+
+<script>
+    document.addEventListener('DOMContentLoaded', (event) => {
```
tcouch, line 117: A simpler approach would have been to embrace the jQuery:

```
const collapsibleElements = $('.collapse');
collapsibleElements.on('show.bs.collapse', (event) => {
    var btn = $(event.target).parent().find('.collapse-button');
   // etc
}
collapsibleElements.on('hide.bs.collapse', (event) => {
  // etc
}
```

tcouch: The "Expand All" button also needs to have it's contents toggled to "Collapse All" if clicked.

src/rard/templates/research/partials/ordered_work_area.html
```diff
@@ -31,10 +33,12 @@
                     <div>
                         {% if work.unknown == True and not flinks and not tlinks and not alinks %}
                         {% elif work.unknown == True %}
-                        <b class='my-3 mr-3'><a href='{% url "work:detail" work.pk %}'>{{ work.name }}</a></b>
+                        <b class='my-3 mr-1'><a href='{% url "work:detail" work.pk %}' >{{ work.name }}</a></b>
```
tcouch, line 36: In #381 you suggested it might be nice to have a little badge with the number of links in the collapsed section. I agree.

src/rard/templates/research/partials/ordered_work_area.html
```diff
@@ -31,10 +33,12 @@
...
+                        <button type='button' class="btn btn-text text-primary pl-2 collapse-button" data-toggle="collapse" data-target="#list_{{work.pk}}" role="button" aria-expanded="false" aria-controls="list_{{work.pk}}"><i class='fas fa-chevron-down'><small class='ml-1 text-muted'>{% trans "Expand" %}</small></i></button>
```
tcouch: Does `btn-text` exist in Bootstrap4? How about `btn-light`? And possibly with `btn-sm` to make it the same size as the other buttons in that row when in edit mode?

tcouch (approved):
Definitely makes it easier to navigate the page.

src/rard/templates/research/partials/ordered_work_area.html
acholyn: I had tried it with the event listener but couldn't get it to work the way I did it (with addEventListener), does this way work? Also if it does, if you open one section and not another and then click expand all, does it toggle them or does it make them all open?

src/rard/templates/research/partials/ordered_work_area.html
acholyn: ah yes, thanks for the reminder!

src/rard/templates/research/partials/ordered_work_area.html
acholyn: I think `btn-text` does exist since it hasn't done the outline or filled it in....but looking at the docs it seems actually to be `btn-link` 🤔 I'll update it to be doubly sure

src/rard/templates/research/partials/ordered_work_area.html
acholyn: have also tried this and it doesn't seem to work on my machine :/

tcouch: It appears that it's a simple toggle so if some sections are already open when a user clicks "Expand All", they would close while the other sections open.

A way to fix that might be to write a custom function to add the "show" class to all the collapsible sections rather than relying on Bootstrap's multi-collapse approach

src/rard/templates/research/partials/ordered_work_area.html
acholyn: image

do we like it before or after the expand button?

tcouch: Before. But... what if the number was *in* the expand button? Like, rather than having the chevron and the word Expand, make it number + chevron 🤯 ?

acholyn: image

I suppose the problem with that is the number then gets quite small but I can probably style it

tcouch: image

acholyn: image
image

I'm liking the outlined version but if you much prefer the filled in then Let me know

tcouch: Oh yeah, outlined is nice

PR #497, opened 2026-01-23
Updated docker and Python dependencies

tim-band:
It took a bit of effort to get the docker compose working. Here are the changes I needed to make.

- slim-buster no longer available, updated to trixie
- updated Python dependencies to work on trixie
- removed obsolete Dockerfile version:
- Dockerfile pull_policy to prevent docker from trying to pull from docker hub
- external postgres port should not clash with local postgres
- docker volume flags :z and :Z are ignored (and attract warnings)
- updated the README


README.md
```diff
@@ -258,26 +226,26 @@ If you have made changes to your models and have not generated a migration file
 
 NB In our case we would run the suggested `manage.py` commands via the Docker container in the way shown above, i.e.
 
-```docker-compose -f local.yml run django python manage.py makemigrations```
+```./manage makemigrations```
```
tcouch, line 242: I think sticking with the container route makes more sense since there are a few env vars that need to be present.

README.md
```diff
@@ -158,27 +126,27 @@ e.g.:
 
 (for example `createsuperuser`)
 
-```docker-compose -f local.yml run django python manage.py <command> [arguments]```
+```./manage <command> [arguments]```
```
tcouch, line 140: Any particular reason for not using the container route?

src/requirements/production.txt
```diff
@@ -3,7 +3,7 @@
 -r base.txt
 
 gunicorn==20.0.4  # https://github.com/benoitc/gunicorn
-psycopg2==2.8.6  # https://github.com/psycopg/psycopg2
+psycopg2==2.9.11  # https://github.com/psycopg/psycopg2
```
tcouch, line 6: Is it worth considering an upgrade to psycopg3? Or is that too complicated?

src/local.yml
```diff
@@ -33,15 +32,16 @@ services:
...
     volumes:
-      - local_postgres_data:/var/lib/postgresql/data:Z
-      - local_postgres_data_backups:/backups:z
+      - local_postgres_data:/var/lib/postgresql/data
+      - local_postgres_data_backups:/backups
     env_file:
       - ./.envs/.local/.postgres
     ports:
-      - "5432:5432"
+      - "15432:5432"
```
tcouch: Is this to avoid conflicts with other local postgres instances?

tim-band: Yes, exactly

tim-band: I had just this problem and it wasn't obvious what the problem was!

README.md
tim-band: It makes files owned by root. But I suppose this formulation is bad too.
Hmmm. Why did this work? What should we do?

tim-band: Oh no, wait! That's not what I was doing!
I had better update the documentation to explain what's actually going on.

src/requirements/production.txt
tim-band: Just tried it. It leads eventually to requiring PostgreSQL 14, which would be a bit of a pain to migrate to.

src/local.yml
tim-band: I actually wondered whether exposing the port is a good idea at all. Does anybody actually use the exposed port or should we delete it?

tcouch: Since this project was originally based on cookiecutter django I had a look to see what the latest version does and it doesn't appear to use ports anymore:
https://github.com/cookiecutter/cookiecutter-django/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/docker-compose.local.yml

src/requirements/production.txt
tcouch: Fair enough

README.md
tcouch: Re the files owned by root problem, could we incorporate this solution into your manage scripts?
`docker run --rm -it --user="$(id -u):$(id -g)" image command`
From this issue: https://github.com/cookiecutter/cookiecutter-django/issues/2284

tcouch (approved):
Approved. Looks good to me.

README.md
tim-band: Nice idea!

src/local.yml
tim-band: The port doesn't need to be exposed for another container in the compose to use it; you only need it if you want to communicate with postgres from the outside. I take it from your answer that we don't care about that, so I'll remove it.

PR #500, opened 2026-03-11
search.py documentation, comments and type annotations

tim-band:
I was trying to do some refactoring ready to do some actual optimization, but this code was too hard for me to understand.

Here is the result of this effort. There is a little refactoring to do with adding snippets, but most of it is documentation and type annotations.

Is this something you can get behind?


tcouch (approved):
Definitely an improvement. Well done! :-)

Issue #494, opened 2025-11-25
Speed up loading Varro


tcouch:
The detail page for the antiquarian [Varro](https://www.ucl.ac.uk/history/frrant/antiquarian/21/) is a particular problem, taking 20s or so on a good day to load. This would most likely be due to the many database queries that run in order to fetch all of the works and all of their fragments.

First port of call would probably be looking at improving efficiency by prefetching related data, but lazy loading accordion content with htmx would also be an option.

(might interfere with staticification?)

Issue #493, opened 2025-11-25
Make antiquarian name matches weigh more in search results


tcouch:
Search for Marcus Terentius Varro should put Varro top

tcouch:
Because we search antiquarians' name, introduction, and re_code fields, even a search limited to just Antiquarians wouldn't put Varro first because of the high number of other Antiquarians mentioning Varro in their introductions. We need some sort of weighting for search results such that a match with name ranks higher than a match anywhere else.

Where the search content type isn't limited to just Antiquarians we might want to think about setting a default ranking; e.g. Antiquarians > Works > Fragments. As a rule of thumb: content types with fewer items should come before those with more.

PR #501, opened 2026-05-11
Speed up search

tim-band:
Speeds up search.
* removed regexing every single original text in the DB at the start of each query
* folded original text denormalization added, so it doesn't have to be done each query
* migration 0056 made reversible (easy fix)
* weakened a couple of tests (sorry) -- but I don't think they were that important. For example varro will match "var,ro".

There is a little problem with this in that the snippet now has all folding applied, rather than just the folding that was necessary to get this search done, which can look a bit weird sometimes. We could change the snippet finding code to look at the plain (rather than folded) text, in which case we'll see the unfolded text, but only if the query matches without folding (which is I think how it works right now anyway)

This speeds up the query for "Varro" from 17-18 seconds down to about 11.


src/rard/utils/text_processors.py
```diff
@@ -23,3 +23,59 @@ def make_plain_text(content):
...
+)
+
+
+def fold_latin(content: str) -> str:
+    for fold_to, fold_from in rard_folds:
+        content = content.replace(fold_from, fold_to)
+    return content
+
+
+def fold_latin_and_remove_punctuation(content: str) -> str:
```
tcouch, line 78: Does make_plain_text not already remove punctuation? No harm in doing it twice I suppose.

tcouch (approved):
This looks great. The only thing is this doesn't actually address issue #494 as that's to do with the speed with which the Antiquarian page itself loads.

The priority for search is issue #493 which might be confusing things. At the moment a search for "Varro" puts the actually Antiquarian page near the end. So the order in which content types are returned in search results needs fixing. Could be done in another PR though if you want to merge this.

PR #504, opened 2026-06-23
Fix for #493

### LINKED ISSUE #493: Make antiquarian name matches weigh more in search results

tim-band:
Antiquarian name matches come higher in search results.

I have annotated each object in the results set with a `weight` integer. Normally this is 1, but I have a new matcher function `match_heavy` that sets it to 100 (giving us plenty of room for intermediate weights; although there's no problem renumbering at any time I suppose).

Fixes #493

### DIFF
--- a/src/rard/research/tests/views/test_search.py
+++ b/src/rard/research/tests/views/test_search.py
@@ -1,6 +1,7 @@
 import pytest
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
+from django.http.request import HttpRequest
 from django.test import RequestFactory, TestCase
 from django.urls import reverse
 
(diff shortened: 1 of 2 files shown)


src/rard/research/tests/views/test_search.py
```diff
@@ -403,6 +404,59 @@ def do_search(search_function, keywords):
...
+        expected = {a1, a2, w1, a1.unknown_work, cw, f1, tm1, f3, af1}
+        self.assertSetEqual(set(results), expected)
+        # But the antiquarian that matched in the name comes first
+        self.assertEqual(results[0], a1)
+        # But when finding the aquarian by a word in its introduction...
+        results2 = self.search_all("two")
+        # All these objects are returned
+        self.assertSetEqual(set(results2), {a2, w2, f2})
+        # And the antiquarian is not first
+        self.assertNotEqual(results2[0], a2)
```
tcouch, line 453: Nice explanation, but I don't think you'd want to assert that antiquarians should explicitly not come first if the match is to its intro instead of its name. If it doesn't matter to us what the order is, I wouldn't test for it.

src/rard/research/views/search.py
```diff
@@ -320,6 +322,32 @@ def match(
                 add_snippet=add_snippet,
             )
 
+        def match_heavy(
+            self, query_set: QuerySet, query_string: str, add_snippet: bool = False
+        ) -> QuerySet:
+            """
+            Get the queryset for matching one type of objects, without Latin folding,
+            with extra weight (so they appear at the top of the)
```
tcouch, line 330: At the top of the...? Please, I need to know! :-)

src/rard/research/views/search.py
```diff
@@ -488,9 +516,9 @@ def antiquarian_search(
         """
         qs = cls.get_filtered_model_qs(Antiquarian, ant_filter=ant_filter)
         search_fields = [
-            ("name", terms.match),
+            ("name", terms.match_heavy),
             ("plain_introduction", terms.match),
-            ("re_code", terms.match),
+            ("re_code", terms.match_heavy),
```
tcouch, line 521: I could take or leave re_code being match_heavy - as far as I can tell it doesn't tend to contain anything the name doesn't have. Although that raises the question of why it's even a search field anyway.

tcouch (approved):
Looks good to me.

acholyn (commented):
can this be merged or closed?

PR #505, opened 2026-06-25
492/antiquarian bibliographic work

### LINKED ISSUE #492: Add Biography/About the Antiquarian [work]
With Unknown works, some of their contents are genuinely unknown, whilst others were attached to the antiquarian directly and are about the antiquarian. Having a biography work will help distinguish these.

tcouch:
closes #492 

* Updates the Work model with a bibliographic boolean flag.
* As per unknown works, an Antiquarian on save signal triggers a check to see if it has a bibliographic work and creates one if not
* Also checks for duplicate biographical works and combines them into one if so (shouldn't happen anyway)
* Fixes bug spotted in unknown work's remove duplicates method which would have stolen links from other, non-unknown works
* Incorporates bibliographic work into ordering logic so it is always the second to last work before unknown
* Updates/fixes the detail template so it does not appear reorderable
* Migration includes a step to add an empty biographical work to all existing Antiquarians


src/rard/research/migrations/0076_add_bibliographic_works.py
```diff
@@ -0,0 +1,45 @@
...
+            model_name='historicalwork',
+            name='bibliographic',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AddField(
+            model_name='work',
+            name='bibliographic',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.RunPython(add_bibliographic_works, reverse_code=migrations.RunPython.noop),
```
tim-band, line 44: Nice!

src/rard/research/tests/models/test_work.py
```diff
@@ -394,6 +400,36 @@ def test_collate_unknown(self):
...
+        self.assertEqual(
+            w.antiquarian_work_fragmentlinks.filter(book=keep_book).count(), 1
+        )
+        self.assertEqual(
+            w.antiquarian_work_fragmentlinks.filter(book=additional_unknown).count(), 1
+        )
+
+        collate_unknown(w)
+
+        self.assertEqual(w.book_set.filter(unknown=True).count(), 1)
```
tim-band, line 425: I'm really not super clear about what this is testing. Are we saying that `additional_unknown` has gone missing? We could test for that explicitly... anyway it feels like we need a comment explaining what these assertions are actually asserting.

src/rard/templates/research/partials/ordered_work_area.html
```diff
@@ -19,93 +19,92 @@
...
-                        <button type='button' class="btn btn-outline-primary btn-sm pl-2 badge badge-pill collapse-button sans-serif" data-toggle="collapse" data-target="#list_{{work.pk}}" role="button" aria-expanded="false" aria-controls="list_{{work.pk}}">{{ tlinks.count|add:flinks.count|add:alinks.count }}<i class='fas ml-2 fa-chevron-down'>
-                        </i></button>
-
-                        {% else %}
-                        <b class='my-3 mr-1'><a href='{% url "work:detail" work.pk %}'>{{ work.name }}</a></b>
-                        <button type='button' class="btn btn-outline-primary collapse-button sans-serif badge badge-pill" data-toggle="collapse" data-target="#list_{{work.pk}}" role="button" aria-expanded="false" aria-controls="list_{{work.pk}}">{{ tlinks.count|add:flinks.count|add:alinks.count }}<i class='fas ml-2 fa-chevron-down'>
-                        </i></button>
-
-                        {% if can_edit and has_object_lock %}
+        {% with tlinks=object|testimonium_links_for_work:work %}
```
tim-band, line 24: nice

tim-band (approved):
Yeah go on then.

